PR #6, opened 2021-02-03
Add '+' char to IssueId validation

jiridostal:
'+' is a valid character to be contained in package names and versions. 
Some examples:
Package name: 'libstdc++'
Package with module: qemu-guest-agent-15:4.2.0-34.module+el8.3.0+8829+e7a0a3ea.1.x86_64

Currently, remediation fails on API validation due to this.

### DIFF
--- a/src/api/openapi.yaml
+++ b/src/api/openapi.yaml
@@ -713,7 +713,7 @@ components:
 
     IssueId:
       type: string
-      pattern: ^(advisor|vulnerabilities|ssg|test|patch-advisory|patch-package):[\w\d_|:\.-]+$
+      pattern: ^(advisor|vulnerabilities|ssg|test|patch-advisory|patch-package):[\w\d_|:\.+-]+$
       example: advisor:CVE_2017_6074_kernel|KERNEL_CVE_2017_6074
 
     RemediationAutoReboot:
--- a/src/connectors/vmaas/mock.js
+++ b/src/connectors/vmaas/mock.js
@@ -50,6 +50,15 @@ const PACKAGES = {
     },
     'systemd-239-13.el8_0.5.x86_64': {
         description: 'systemd is a system and service manager that runs as PID 1 and starts.'
+    },
+    'libstdc++-8.3.1-5.1.el8.x86_64': {
+        description: 'GNU Standard C++ Library'
+    },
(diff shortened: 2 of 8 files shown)


jiridostal:
@aleccohan could you please have a look? It does not allow Patch to select all packages on the system (which will most likely contain a package with '+' in it) and remediate it.

aleccohan (changes requested):
@jiridostal 2 quick things.

1. Wrt the placement of this regex pattern this is not actually the place where the issue_id is checked.  We do that here --> https://github.com/RedHatInsights/insights-remediations/blob/master/src/util/identifiers.js#L4.  Additionally would you be able to add some unit tests for issues that look like that in here --> https://github.com/RedHatInsights/insights-remediations/blob/master/src/util/identifiers.unit.js?

2.  Did you re-clone this repo before pushing this commit?  I just recently shifted Insights-remediation to be a public repo and as such I had to scrub the commit history.  Just want to make sure that none of the old changes get pushed back up!

jiridostal:
1. Fixed the other place and added test cases for Patch
2. I've rebased these two commits on top of this repo, so that shouldn't be a problem.

aleccohan (changes requested):
Hey Jiri!  So those changes are looking good now.  However I think that (based on the format of some of these issues) the PKG_PATTERN regex defined --> https://github.com/RedHatInsights/insights-remediations/blob/master/src/issues/index.js#L10 will also fail for issues of this format.  

Additionally, this PKG_PATTERN is used in the default package templates stored in remediations as input sanitation.  Could you update the Pattern here as well? 
https://github.com/RedHatInsights/insights-remediations/blob/master/src/templates/static/patchman/packages.yml

Josca:
@aleccohan thanks, I've updated PKG_PATTERN regex and added some tests. It should be fixed now.

aleccohan:
@Josca Thanks!  This looks great.  Just one more quick thing before it is merged.  Would you be able to update the package regex in the default package template to match the updated package regex?
i.e.  https://github.com/RedHatInsights/insights-remediations/blob/master/src/templates/static/patchman/packages.yml#L9

Once that's in then I'll go ahead and approve it!

Josca:
@aleccohan sure, I've just did it in separate PR as this is @jiridostal 's one.

Josca:
@jiridostal you can close that. Updates were merged in #10 .

PR #12, opened 2021-02-12
RHCLOUD-11047: local connection support for split playbooks

aleccohan:
Added local connection support for arbitrarily split playbooks.

Main features:
1.  Added the _localhost_ query parameter to the **/playbooks** method.
2. Added new rebootLocalHost template to encompass the changes made for reboot on local host. 
3. Added 1min delay to reboot if _localhost_ flag has been declared.
4. Removed post_check_in step from reboot if _localhost_ flag is declared.
5. Added command **insights-client** to reboot as a checkup after the reboot is completed.
6. Added additional unit-tests for this new feature
 
    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/templates/static/special/rebootLocalHost.yml
```diff
@@ -0,0 +1,17 @@
...
+    insights_signature_exclude: "/hosts"
+  become: True
+  gather_facts: False
+  tasks:
+    - when:
+        - insights_needs_reboot is defined
+        - insights_needs_reboot
+      block:
+        - name: Reboot system
+          reboot:
```
jharting, line 15: I don't think this is going to work for localhost as the machine running the playbook is the one that gets rebooted. Have you tested this?

aleccohan: @jharting I have, and I now see that this module does not run for the reason you illustrate above.  Looks like I'll change it back to:
> sleep 2 && shutdown -r +1

aleccohan: I've replaced the logic with the one I stated above.  Additionally based on the conversation with Link I added the "insights-client post reboot" step to create the file designated in this unit file PR:
https://github.com/RedHatInsights/insights-client/pull/138/files

src/templates/static/special/rebootLocalHost.yml
```diff
@@ -0,0 +1,21 @@
...
+        - insights_needs_reboot is defined
+        - insights_needs_reboot
+      block:
+        - name: Reboot system
+          shell: sleep 2 && shutdown -r +1 now "Ansible triggered reboot"
+          async: 1
+          poll: 0
+          ignore_errors: true
+        - name: Schedule insights-client after boot
+          shell: touch /.run_insights_client_next_boot
```
jharting, line 20: Using literaly `/.run_insights_client_next_boot` seems odd. Is there a more appropriate place to put this? E.g. insights-client configuration directory?

jharting, line 20: might be better to use ansible built in file module to do the touching instead of calling shell

src/templates/static/special/rebootLocalHost.yml
```diff
@@ -0,0 +1,21 @@
...
+    - when:
+        - insights_needs_reboot is defined
+        - insights_needs_reboot
+      block:
+        - name: Reboot system
+          shell: sleep 2 && shutdown -r +1 now "Ansible triggered reboot"
+          async: 1
+          poll: 0
+          ignore_errors: true
+        - name: Schedule insights-client after boot
```
jharting, line 19: Doing things in this order creates a potential for race condition (host gets rebooted before the file gets touched). It is very unlikely but it may still be better to trigger reboot after everything has been set up.

src/templates/static/special/rebootLocalHost.yml
```diff
@@ -0,0 +1,21 @@
...
+  become: True
+  gather_facts: False
+  tasks:
+    - when:
+        - insights_needs_reboot is defined
+        - insights_needs_reboot
+      block:
+        - name: Reboot system
+          shell: sleep 2 && shutdown -r +1 now "Ansible triggered reboot"
+          async: 1
```
jharting, line 16: IIRC async/poll are needed when the host in question disconnects (do to reboot). In this case we're scheduling it to happen after the playbook finishes and therefore they should not be needed. Same for `sleep 2`

src/templates/static/special/rebootLocalHost.yml
aleccohan: This is a good question.  I'll reach out to Link about it to see what I can come up with.

aleccohan: @jharting After some discussion with Link the more appropriate place for this file would have to be **/etc/insights-client**.  I'll make that change in this next iteration.

jharting (approved):

PR #14, opened 2021-02-25
RHCLOUD-12357: Allow cert-auth on /playbooks endpoint

aleccohan:
This PR allows for the use of cert-auth when calling the /playbooks endpoint.  This will allow the client playbook-runner to get playbook yaml as long as the owner_id in the cert-auth header matches the owner_id in the hosts system_profile.
    
    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/routes.js
```diff
@@ -68,8 +68,9 @@ module.exports = async function (app) {
     require('./diagnosis/routes')(v1);
     require('./diagnosis/routes')(legacy);
     require('./whoami/routes')(v1);
+    require('./remediations/playbook/routes')(v1);
 
-    // diagnosis and whoami are the only path that accepts cert auth
+    // diagnosis, whoami, and remediaitons/playbooks are the only path that accepts cert auth
```
Jason-RH, line 73: small typo here

src/remediations/remediations.queries.js
```diff
@@ -168,8 +168,8 @@ exports.loadDetails = async function (account_number, created_by, rows) {
     return rows.map(row => _.assign(byId[row.id].toJSON(), row));
 };
 
-exports.get = function (id, account_number, created_by) {
-    return db.remediation.findOne({
+exports.get = function (id, account_number, created_by =  null) {
```
Jason-RH, line 171: super minor but did you intend to put two spaces before 'null' here?

Jason-RH (approved):
Seems reasonable to me :+1:

src/remediations/controller.read.js
```diff
@@ -27,6 +27,33 @@ const catchErrorCode = (code, fn) => e => {
     throw e;
 };
 
+async function fetchRemediation (req) {
+    const id = req.params.id;
+    const hosts = req.query.hosts;
```
jharting, line 32: what happens if no `hosts` query param is provided?

src/remediations/controller.read.js
```diff
@@ -27,6 +27,33 @@ const catchErrorCode = (code, fn) => e => {
...
+
+    if (_.isUndefined(req.user)) {
+        const batchProfileInfo = await inventory.getSystemProfileBatch(hosts); // TODO: change batchProfileInfo var name
+
+        if (_.isEmpty(batchProfileInfo)) {
+            return [];
```
jharting, line 38: why `[]`?

src/remediations/controller.read.js
```diff
@@ -27,6 +27,33 @@ const catchErrorCode = (code, fn) => e => {
...
+            if (_.has(batchProfileInfo, host)) {
+                // eslint-disable-next-line security/detect-object-injection
+                const ownerId = batchProfileInfo[host].system_profile.owner_id;
+                if (!_.isEqual(req.identity.system.cn, ownerId)) {
+                    throw errors.unauthorizedGeneration(req.identity.account_number);
+                }
+            }
+        });
+
+        return await queries.get(id, req.identity.account_number);
```
jharting, line 51: since this function is async you can return promise (no need to await)

src/remediations/playbook/routes.js
```diff
@@ -0,0 +1,15 @@
...
+const openapi = require('../../middleware/openapi');
+const rbac = require('../../middleware/rbac');
+
+const rbacRead = rbac('remediations:remediation:read');
+
+module.exports = function (router) {
+    router.get(
+        '/remediations/:id/playbook',
+        openapi('getRemediationPlaybook'),
+        rbacRead,
```
jharting, line 13: I am not sure if, when cert identity is used, the rbac middleware is actually going to let the request through.

src/errors.js
```diff
@@ -163,6 +163,9 @@ exports.noExecutors = remediation =>
 exports.unknownExclude = excluded =>
     new exports.BadRequest('UNKNOWN_EXCLUDE', `Excluded Executor [${excluded}] not found in list of identified executors`);
 
+exports.unauthorizedGeneration = account =>
+    new exports.Forbidden(`Account ${account} is unauthorized to access this playbook`);
```
lphiri, line 167: This error will be confusing.  Its not the the account that does not have the right permission, but rather the certificate being presented. Perhaps change this to accept `req.identity.system.cn` instead and change the error message appropriately?

src/remediations/controller.read.js
```diff
@@ -27,6 +27,33 @@ const catchErrorCode = (code, fn) => e => {
...
+        if (_.isEmpty(batchProfileInfo)) {
+            return [];
+        }
+
+        hosts.forEach(host => {
+            if (_.has(batchProfileInfo, host)) {
+                // eslint-disable-next-line security/detect-object-injection
+                const ownerId = batchProfileInfo[host].system_profile.owner_id;
+                if (!_.isEqual(req.identity.system.cn, ownerId)) {
+                    throw errors.unauthorizedGeneration(req.identity.account_number);
```
lphiri, line 46: See comment in  `unauthorizedGeneration` function - pass the CN instead?

src/remediations/controller.read.js
aleccohan: @jharting At the moment should no "hosts" param be provided and the /playbook method be called with cert-auth the operation returns a 404 error due too the owner_id check not being able to run.  Would it make more sense for this to be a 400?

src/errors.js
aleccohan: Will do!

src/remediations/controller.read.js
aleccohan: Ah, yes this is redundant, I'll return batchProfileInfo since it's already empty.

aleccohan: This was meant to trigger the "404" error in the playbooks function by returning an "empty" remediation but It would probably make more sense to throw that 404 error here.

jharting: Is it going to 404, though? Walking through the code it seems to me that it's going to choke on `remediation.toJSON()` instead.

src/remediations/controller.read.js
jharting: I am wondering if in that case we should return the playbook but only as long as the only host used in the playbook is the one that matches the cert. Or perhaps automatically return a slice that matches the given cert?

aleccohan:
@lphiri will do!  I'll go ahead and add a comment to the RBAC middleware to make the usecase more explicit.

jharting (approved):

PR #18, opened 2021-03-09
RHCLOUD-12760: Modify Remediations to include RHC compatability

aleccohan:
The purpose of this PR is to modify the current Remediations FiFi pipeline to send RHC registered systems in a work request to the playbook-dispatcher without disturbing the pipeline that is already in place.  

To that end, the way that the pipeline works now is that instead of having a sole executor at the end of the "Preflight Check" that are not runnable due to being "Direct Connect" we now comb through these systems to see if they have an existing "rhc_client_id" if they do they are added to an executor that is marked as "connected" with the type "RHC".  All other systems (without an existing "rhc_client_id" are put in a seperate executor which will show up with an error stating that they cannot be run with FiFi.  

Additionally with these changes I had to modify one of the existing features, specifically the ability to exclude executors from being run by FiFi.  You are now able to add the executor "RHC" to that list which will exclude any system that is RHC enabled from being run.

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/remediations/fifi.integration.js
+++ b/src/remediations/fifi.integration.js
@@ -64,14 +64,14 @@ describe('FiFi', function () {
             .set(auth.fifi)
             .expect(200);
 
-            headers.etag.should.equal('"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"');
+            headers.etag.should.equal('"1172-aL66zBR3zzAbkQBquBARuVlS5w0"');
         });
 
         test('304s on ETag match', async () => {
             await request
             .get('/v1/remediations/0ecb5db7-2f1a-441b-8220-e5ce45066f50/connection_status?pretty')
             .set(auth.fifi)
-            .set('if-none-match', '"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"')
+            .set('if-none-match', '"1172-aL66zBR3zzAbkQBquBARuVlS5w0"')
             .expect(304);
         });
     });
@@ -778,14 +778,14 @@ describe('FiFi', function () {
                 await request
                 .post('/v1/remediations/0ecb5db7-2f1a-441b-8220-e5ce45066f50/playbook_runs')
                 .set(auth.fifi)
-                .set('if-match', '"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"')
+                .set('if-match', '"1172-aL66zBR3zzAbkQBquBARuVlS5w0"')
                 .expect(201);
             });
 
             test('400 post playbook run', async () => {
                 await request
                 .set(auth.fifi)
-                .get('/v1/remediations/66eec356-dd06-4c72-a3b6-ef27d150000/connection_status')
+                .get('/v1/remediations/66eec356-dd06-4c72-a3b6-ef27d150000/playbook_runs')
                 .expect(400);
             });
 
@@ -810,17 +810,17 @@ describe('FiFi', function () {
                 const {headers} = await request
                 .post('/v1/remediations/0ecb5db7-2f1a-441b-8220-e5ce45066f50/playbook_runs?pretty')
                 .set(auth.fifi)
-                .set('if-match', '"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"')
+                .set('if-match', '"1172-aL66zBR3zzAbkQBquBARuVlS5w0"')
                 .expect(201);
 
-                headers.etag.should.equal('"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"');
+                headers.etag.should.equal('"1172-aL66zBR3zzAbkQBquBARuVlS5w0"');
             });
 
             test('201s on ETag match', async () => {
                 await request
                 .post('/v1/remediations/0ecb5db7-2f1a-441b-8220-e5ce45066f50/playbook_runs')
                 .set(auth.fifi)
-                .set('if-match', '"10d1-znP/hi4fEIT3PozsJd+nSvx8Im4"')
+                .set('if-match', '"1172-aL66zBR3zzAbkQBquBARuVlS5w0"')
                 .expect(201);
             });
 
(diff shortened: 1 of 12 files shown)


src/remediations/fifi.js
```diff
@@ -41,17 +49,49 @@ exports.getSatelliteId = function (facts) {
     return null;
 };
 
+function defineRHCEnabledExecutor (satellites) {
+    const lastElement = _.last(satellites);
```
jharting, line 53: how confident are you that the "bucket of non-satellite systems" is always going to be the last element in the list?

src/remediations/fifi.js
```diff
@@ -191,18 +243,25 @@ exports.getConnectionStatus = async function (remediation, account) {
...
+    if (!_.isEmpty(satellites)) {
+        defineRHCEnabledExecutor(satellites);
+    }
+
     const sourceInfo = await sources.getSourceInfo(_(satellites).map('id').filter().value());
 
     _.forEach(satellites, satellite => {
         satellite.source = _.get(sourceInfo, satellite.id, null);
         satellite.receptor = getReceptor(satellite.source);
+        satellite.rhcStatus = fetchRHCStatus(satellite.systems);
```
jharting, line 264: This looks very brittle. Say you have 5 sat-enabled hosts. Should one of them define rhc_client for some reason (e.g. test data, leftover from when the host was registered directly before moving under satellite, etc.) that one host is going to make the entire Satellite as rhc_connected. As a result, remediations will from there on consider the entire executor (Satellite) as RHC and use RHC for running stuff on any of the 5 hosts.

Instead, it should be clearly defined who (Satellite vs RHC) wins and if a host is initially put into a "Satellite bucket" remediations should not under no circumstances convert the whole bucket into RHC.

src/remediations/fifi.js
```diff
@@ -123,32 +163,44 @@ function getName (executor) {
...
-    if (executor.receptorStatus !== 'connected') {
+    if (executor.receptorStatus !== 'connected' && executor.rhcStatus === 'disconnected') {
         return 'disconnected';
     }
 
     return 'connected';
 }
 
+function getType (executor) {
+    if (executor.id && executor.rhcStatus === 'disconnected') {
```
jharting, line 186: I think it would be more robust to explicitly define executor.type = satellite or rhc when creating an executor "bucket" rather than inferring it indirectly from connection status, id, etc.

src/remediations/fifi.js
```diff
@@ -41,17 +49,49 @@ exports.getSatelliteId = function (facts) {
...
+        const nonRhcSystems = _.filter(lastElement.systems, system => _.isUndefined(system.rhc_client));
+
+        lastElement.systems = _.filter(lastElement.systems, system => !_.isUndefined(system.rhc_client));
+
+        satellites.push({id: null, systems: nonRhcSystems});
+    }
+}
+
+function fetchRHCStatus (systems) {
+    let rhcConnectionStatus = 'disconnected';
```
jharting, line 64: Have you considered checking the connection status with cloud-connector similarly to how we do with receptor? The fact that a host has `rhc_client` defined does not necessarily indicate that it is connected right now. No need to make it part of this initial request just something to consider.

src/remediations/fifi.integration.js
```diff
@@ -1166,7 +1166,7 @@ describe('FiFi', function () {
                                 display_name: null
                             }
                         ],
-                        type: 'Satellite',
+                        type: 'satellite',
```
jharting, line 1169: this looks like potentially breaking - e.g. frontend may be expecting the previous value?

src/remediations/fifi.js
```diff
@@ -123,32 +163,44 @@ function getName (executor) {
 }
 
 function getStatus (executor) {
-    if (!executor.id) {
+    if (!executor.id && executor.rhcStatus === 'disconnected') {
```
jharting, line 166: again, would be way cleaner to have
```
if (executor.type == "satellite") {
    // do satellite checks
} else if (executor.type == "RHC") {
    // do RHC check if needed
} else {
    // default
}
```

than indirectly inferring whether the given executor is rhc or not from its status.

jharting (commented):
Looks pretty good. Left a few comments.

src/remediations/fifi.js
aleccohan: Will do!  I was planning on adding in the check to do that after this initial request.

src/remediations/fifi.integration.js
aleccohan: @jharting I checked the Frontend code for fifi and it looks like they're looking for a lowercase 'satellite' so I believe this should be alright.

src/remediations/fifi.js
```diff
@@ -191,18 +245,26 @@ exports.getConnectionStatus = async function (remediation, account) {
...
+    if (!_.isEmpty(satellites)) {
+        defineRHCEnabledExecutor(satellites);
+    }
+
     const sourceInfo = await sources.getSourceInfo(_(satellites).map('id').filter().value());
 
     _.forEach(satellites, satellite => {
         satellite.source = _.get(sourceInfo, satellite.id, null);
         satellite.receptor = getReceptor(satellite.source);
+        satellite.type = getExecutorType(satellite);
```
jharting, line 266: won't this mark even the no-satellite no-rhc "bucket" as type=RHC?

src/remediations/fifi.js
```diff
@@ -41,17 +49,51 @@ exports.getSatelliteId = function (facts) {
...
 };
 
+function defineRHCEnabledExecutor (satellites) {
+    const satlessExecutor = _.find(satellites, satellite => satellite.id === null);
+    if (satlessExecutor) {
+        const partitionedSystems = _.partition(satlessExecutor.systems, system => !_.isUndefined(system.rhc_client));
+
+        satlessExecutor.systems = partitionedSystems[0];
+
+        satellites.push({id: null, systems: partitionedSystems[1]});
```
jharting, line 59: this may be a good place to add `type: RHC`

dehort (approved):
:+1:

PR #22, opened 2021-03-19
RHCLOUD-11976: Update Smart-Management check to be in controllers & only effect "execute" and "pre-flight check"

aleccohan:
Updated the Smart-management middleware to also check each of the systems (as part of the playbook related call) and if at least one of them has the "is_marketplace" flag as true, allow the user to hit the Remediations API.  This functionality has also been "locked" behind an environment variable so that we can turn it on when the time is right.

As a note wrt line 16 of the smartManagement.js file, I wasn't sure if this should immediatly throw a 404 or call next() to allow the logic in the fifi.connector file take care of it.  I went with setting res = 404, but I'm absolutely open to change this since I was unsure of the best practice for handling this situation.

UPDATE 03/24/2021:
Updated this PR per the most recent conversation with Jozef to remove the Smart-managment middleware entirely in favor of a more specialized approach.  Instead of running the Smart-Management check against all FiFi endpoints we will now only run the check againsed the pre-flight check and the "execute" method.  Each other option will be visable to the user and show up as empty.

JIRA: https://issues.redhat.com/browse/RHCLOUD-11976

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/middleware/identity/smartManagement.js
```diff
@@ -1,12 +1,40 @@
...
 
-module.exports = function (req, res, next) {
-    if (!req.entitlements.smart_management) {
-        return next(new errors.Forbidden());
-    }
+module.exports = async function (req, res, next) {
+    try {
+        let rhcConnectedSystems = false;
+        if (config.isMarketplace) {
+            const remediation = await queries.get(req.params.id, req.user.account_number, req.user.username);
```
jharting, line 14: This is a costly check. There should be a fast lane for customers with `req.entitlements.smart_management`

src/middleware/identity/smartManagement.js
```diff
@@ -1,12 +1,40 @@
...
 
-    next();
-};
+            if (!remediation) {
+                res.sendStatus(404);
+            }
+
+            const systemsIds = _(remediation.issues).flatMap('systems').map('system_id').uniq().sort().value();
```
jharting, line 20: It's not clear to me whether a marketplace host in the playbook is what actually unlocks the smart management features. My understanding was that a marketplace host (not necessarily involved in the playbook) is what unlocks it. I am going to find out.

jharting: Btw if we get a chance to bake this logic into the entitlements service we should do so. IMO it's not remediations' responsibility to figure out these oddities.

aleccohan: @jharting Just pushed up the "fast lane" changes for customers with the smart_management entitlement.  Also wrt the first question, I believe that a marketplace host only unlocks smart management features when it's part of a playbook.

jharting (approved):
Looks good to me. We may want to tweak the handling of "neither smart management entitlement not marketplace" so that instead of 403 we show a preflight modal. We can discuss/implement that separately.

PR #25, opened 2021-04-05
RHCLOUD-13324: Update dispatcher-connector with new API endpoints

aleccohan:
Updated playbook dispatcher connector to have additional endpoints for pulling current information about running jobs.  Specificly the 'dispatcher/v1/run_hosts' endpoint and the 'dispatcher/v1/runs' endpoint which will be used / configured with the current FiFi endpoints to fetch information about direct-connect jobs.

JIRA: https://issues.redhat.com/browse/RHCLOUD-13324    

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/connectors/dispatcher/impl.js
+++ b/src/connectors/dispatcher/impl.js
@@ -2,15 +2,21 @@
 
 const _ = require('lodash');
 const URI = require('urijs');
+const qs = require('qs');
+const config = require('../../config');
 const {host, insecure, auth} = require('../../config').dispatcher;
 
 const Connector = require('../Connector');
 const metrics = require('../metrics');
 
+const QSOPTIONS = { encode: false };
+
 module.exports = new class extends Connector {
     constructor () {
         super(module);
         this.postRunRequests = metrics.createConnectorMetric(this.getName(), 'postPlaybookRunRequests');
+        this.fetchRuns = metrics.createConnectorMetric(this.getName(), 'fetchPlaybookRuns');
+        this.fetchRunHosts = metrics.createConnectorMetric(this.getName(), 'fetchPlaybookRunHosts');
     }
 
     async postPlaybookRunRequests (dispatcherWorkRequest) {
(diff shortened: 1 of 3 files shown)


src/connectors/dispatcher/impl.js
```diff
@@ -43,5 +45,65 @@ module.exports = new class extends Connector {
         return result;
     }
 
+    async fetchPlaybookRuns (filter = null, fields = null, sort_by = null) {
+        const uri = new URI(config.platformHostname);
+        uri.path('/api/playbook-dispatcher/v1/runs');
```
Jason-RH, line 50: Does this behave differently than uri.segment() used in the postPlaybookRunRequests function?

Jason-RH (approved):
lgtm!

src/connectors/dispatcher/impl.js
aleccohan: Hey!  Yeah so, IIRC uri.segment() adds an additional `/` infront and behind whatever is passed to it, so I could have done something like:
> uri.segment('api');
> uri.segment('playbook-dispatcher')';
...etc

but I figured this would look cleaner.

src/connectors/dispatcher/impl.js
```diff
@@ -43,5 +45,65 @@ module.exports = new class extends Connector {
...
+        }
+
+        const options = {
+            uri: uri.toString(),
+            method: 'GET',
+            json: true,
+            headers: this.getForwardedHeaders()
+        };
+
+        const result = await this.doHttp (options, false, this.getPlaybookRuns);
```
jharting, line 99: should this operation use its own connector metric?

src/connectors/dispatcher/impl.js
```diff
@@ -43,5 +46,65 @@ module.exports = new class extends Connector {
         return result;
     }
 
+    async fetchPlaybookRuns (filter = null, fields = null, sort_by = null) {
+        const uri = new URI(config.platformHostname);
+        uri.path('/api/playbook-dispatcher/v1/runs');
+
+        if (filter) {
+            uri.addQuery('filter', filter);
```
jharting, line 54: both `filter` and `fields` use LHS bracket syntax (e.g. `filter[labels][service]=remediations`). I don't think you'll be able to specify these like this. I think you'll need to do something like `uri.addQuery('filter[labels][service]', 'remediations')`

Based on https://github.com/medialize/URI.js/issues/86 it looks like urijs does not support this natively. Perhaps a custom encoder could be written, though.

aleccohan: @jharting would it also be possible to create a JSON object like what is shown in the API docs and pass that as the data for the filter parameter?  

i.e.
```
{
  "run": {
    "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
    "labels": {
      "additionalProp1": "string",
      "additionalProp2": "string",
      "additionalProp3": "string"
    },
    "service": "string"
  },
  "status": "running"
}
```

jharting: Based on what I found it looks like urijs does not natively support deepObject query params i.e. you'll need to add logic for converting it.

aleccohan: @jharting I think I found a solution in the `qs` library.  I'll have to do some additional manipulation when I declare the filter and fields variable for each call, but this looks like it will return the correct format to be encoded by URI.js.

jharting (approved):

PR #28, opened 2021-04-15
RHCLOUD-13240: Update playbook_runs methods with RHC logic

aleccohan:
Updated the /playbook_runs && /playbook_runs/systems methods with dispatcher logic.  Now each of the GET methods to fetch information about playbook runs also fetches and combines data from the dispatcher if there are any RHC runs present.  

JIRA: https://issues.redhat.com/browse/RHCLOUD-13240    

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/remediations/fifi.js
```diff
@@ -43,6 +50,187 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+};
+
+exports.combineRuns = function (rhcRuns, remediation, remediation_id, created_by) {
+    if (!rhcRuns) {
+        return remediation.playbook_runs;
+    }
+
+    rhcRuns = formatRHCRuns(rhcRuns);
+
+    if (!remediation) {
```
jharting, line 216: in what situation is `remediation` going to be undefined?

src/remediations/fifi.js
```diff
@@ -43,6 +50,187 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+        _.forEach(rhcRuns, rhcRun => {
+            pushRHCRun(rhcRun, remediation, remediation_id, created_by);
+        });
+    }
+
+    _.forEach(rhcRuns, rhcRun => {
+        const satRun = _.find(remediation.playbook_runs, run => run.id === rhcRun[0].labels['playbook-run']);
+        if (_.isUndefined(satRun)) {
```
jharting, line 224: I did not know you're not storing a playbook_run entry in case of RHC-only runs. Have you considered storing it always?

src/remediations/controller.fifi.js
```diff
@@ -87,11 +87,17 @@ exports.listPlaybookRuns = errors.async(async function (req, res) {
     const {column, asc} = format.parseSort(req.query.sort);
     const {limit, offset} = req.query;
     let remediation = await queries.getPlaybookRuns(req.params.id, req.user.account_number, req.user.username, column, asc);
+    const rhcRuns = await fifi.getRHCRuns();
```
jharting, line 90: This will retrieve the entire history of playbook runs, even for other remediations, right?

src/remediations/controller.fifi.js
```diff
@@ -115,10 +121,17 @@ exports.getRunDetails = errors.async(async function (req, res) {
         req.user.username
     );
 
-    if (!remediation) {
+    const rhcRuns = await fifi.getRHCRuns(req.params.playbook_run_id);
```
jharting, line 124: If you make the db query and request to playbook-dispatcher run concurrently I think this operation will be faster.

src/remediations/fifi.js
aleccohan: @jharting This was a product of not storing a playbook_run entry in the case of a RHC-only runs.  However I plan on changing this with the next push, adding the storage of the playbook_run regardless of the executor type.

src/remediations/fifi.js
```diff
@@ -43,6 +50,149 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+    });
+
+    // Status of executor
+    rhcRuns.status = findRunStatus(rhcRuns);
+}
+
+exports.formatRunHosts = function (rhcRunHosts) {
+    return _.map(rhcRunHosts.data, host => ({
+        system_id: host.id,
+        system_name: 'localhost',
```
jharting, line 106: While technically this is correct as `localhost` is the hostname we use with RHC, I think this will get confusing to the customers if they see a paginated table where all entries are "localhost".

src/remediations/fifi.js
```diff
@@ -526,7 +686,9 @@ exports.createPlaybookRun = async function (status, remediation, username, exclu
         const rhcRequests = splitRequests.RHC[0]; // There should only ever be one
         await dispatchRHCRequests(rhcRequests, playbook_run_id);
 
-        // TODO: Store responses??
+        if (_.isEmpty(splitRequests.satellite)) {
```
jharting, line 689: why the isEmpty() check?

src/remediations/fifi.js
```diff
@@ -43,6 +50,149 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+exports.combineHosts = function (rhcRunHosts, systems) {
+    rhcRunHosts = exports.formatRunHosts(rhcRunHosts);
+
+    _.forEach(rhcRunHosts, host => {
+        pushRHCSystem(host, systems);
+    });
+};
+
+exports.combineRuns = async function (remediation) {
+    for (const satRun of remediation.playbook_runs) {
```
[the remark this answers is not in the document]
aleccohan: @jharting Based on what needs to be returned (at least for the "get all playbook runs" method) I'm not sure if there's a way to identify which one is an RHC run before sending the request to the dispatcher since it's possible that each "run" has RHC systems included.

src/remediations/fifi.js
```diff
@@ -43,6 +50,149 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+        const rhcRuns = await exports.getRHCRuns(satRun.id);
+
+        if (rhcRuns) {
+            formatRHCRuns(rhcRuns);
```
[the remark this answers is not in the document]
aleccohan: @jharting yes I believe so, each "playbook_run" _can_ have associated RHC systems this condition can be true for each run that are then subsiquently shown as the "Direct Conneced" executor.

src/remediations/fifi.js
aleccohan: this is for the situation such that there are no satellite runs defined.  When that is the case we store the RHC only run as just the "playbook_run" in the remediations db.

src/remediations/fifi.js
aleccohan: This makes sense.  I'll swap it to something like the "host.id" instead

aleccohan: @jharting once the inventory "link" PR is in I'll swap the 'system_name' to be the inventory host_id returned in links.  I'll make a subsiquent PR for that change.

src/remediations/controller.fifi.js
```diff
@@ -161,22 +176,41 @@ exports.getSystems = errors.async(async function (req, res) {
         }
     }
 
-    const formatted = format.playbookSystems(systems, count);
+    // Pagination
+    const total = fifi.getListSize(systems);
+    systems = await fifi.pagination(systems, total, limit, offset);
```
jharting, line 181: the function is not async --> no need to await

src/remediations/controller.fifi.js
```diff
@@ -161,22 +176,41 @@ exports.getSystems = errors.async(async function (req, res) {
...
+
+    if (offset >= Math.max(total, 1)) {
```
jharting, line 183: might be better to do the check first before using the value in a call to `fifi.pagination()`

src/remediations/fifi.js
```diff
@@ -43,6 +50,150 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+    });
+
+    // Status of executor
+    rhcRuns.status = findRunStatus(rhcRuns);
+}
+
+exports.formatRunHosts = function (rhcRunHosts) {
+    return _.map(rhcRunHosts.data, host => ({
+        system_id: host.id,
+        system_name: host.id,
```
jharting, line 107: I guess that's good enough for now but overall I doubt the uuid is going to be useful for users. We should try to get the host name here somehow.

src/remediations/fifi.js
```diff
@@ -43,6 +50,150 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+        status: host.status,
+        updated_at: host.updated_at,
+        playbook_run_executor_id: exports.generateUuid()
```
jharting, line 110: Not a big fan of this value changing every time the customer refreshes. What about using the remediations run id here instead?

src/remediations/fifi.js
```diff
@@ -43,6 +50,150 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
...
+        system_name: host.id,
+        status: host.status,
+        updated_at: host.updated_at,
+        console: details.data[0].stdout,
+        executor_id: exports.generateUuid()
+    };
+}
+
+function pushRHCSystem (host, systems) {
+    systems.push({
```
jharting, line 126: is this different than doing `systems.push(host)` ?

src/remediations/fifi.js
aleccohan: will do!  I'll make a note to update this after the initial implementation is out.

jharting (approved):

PR #31, opened 2021-04-19
RHCLOUD-13446: Add Config Manager Check to remediations execute & preflight logic

aleccohan:
Added Config Manager check to remediations execute & pre-flight check logic.

JIRA: https://issues.redhat.com/browse/RHCLOUD-13502
   
    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/remediations/__snapshots__/fifi.integration.js.snap
+++ b/src/remediations/__snapshots__/fifi.integration.js.snap
@@ -75,6 +75,148 @@ exports[`FiFi connection status get connection status with false smartManagement
 }"
 `;
 
+exports[`FiFi connection status get connection status with false smartManagment but not configured with config manager 1`] = `
+"{
+    \\"meta\\": {
+        \\"count\\": 8,
+        \\"total\\": 8
+    },
+    \\"data\\": [
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": \\"01bf542e-6092-485c-ba04-c656d77f988a\\",
+            \\"executor_type\\": \\"satellite\\",
+            \\"executor_name\\": \\"Satellite 01bf542e-6092-485c-ba04-c656d77f988a\\",
+            \\"system_count\\": 4,
+            \\"connection_status\\": \\"no_smart_management\\"
+        },
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": \\"409dd231-6297-43a6-a726-5ce56923d624\\",
+            \\"executor_type\\": \\"satellite\\",
+            \\"executor_name\\": \\"Satellite 409dd231-6297-43a6-a726-5ce56923d624\\",
+            \\"system_count\\": 6,
+            \\"connection_status\\": \\"no_smart_management\\"
+        },
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": \\"63142926-46a5-498b-9614-01f2f66fd40b\\",
+            \\"executor_type\\": \\"satellite\\",
+            \\"executor_name\\": \\"Satellite 63142926-46a5-498b-9614-01f2f66fd40b\\",
+            \\"system_count\\": 1,
+            \\"connection_status\\": \\"no_smart_management\\"
+        },
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": \\"722ec903-f4b5-4b1f-9c2f-23fc7b0ba390\\",
+            \\"executor_type\\": \\"satellite\\",
+            \\"executor_name\\": \\"Satellite 722ec903-f4b5-4b1f-9c2f-23fc7b0ba390\\",
+            \\"system_count\\": 3,
+            \\"connection_status\\": \\"no_smart_management\\"
+        },
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": \\"72f44b25-64a7-4ee7-a94e-3beed9393972\\",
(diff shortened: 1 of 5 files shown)


src/remediations/fifi.js
```diff
@@ -37,6 +39,12 @@ exports.checkSmartManagement = async function (remediation, smart_management) {
         return false;
     }
 
+    const rhcStates = await configManager.getCurrentState();
```
jharting, line 42: As this is unrelated to smart management it would be a bit cleaner either to do the check outside of the function or rename the function to have a more general name. Also, since this is a API call there is a potential to optimize performance by making the call in parallel with other slow operations (API calls, db queries, ...).

Jason-RH (approved):

src/remediations/controller.fifi.js
```diff
@@ -26,7 +26,18 @@ exports.connection_status = errors.async(async function (req, res) {
         throw new errors.Forbidden();
     }
 
-    const status = await fifi.getConnectionStatus(remediation, req.identity.account_number, req.entitlements.smart_management);
+    const rhcEnabled = await fifi.checkRhcEnabled();
+
+    if (!rhcEnabled && !req.entitlements.smart_management) {
```
jharting, line 31: Is `&& !req.entitlements.smart_management` there because you can only enable remediations in RHC if you have smart management?

jharting: Well either way I think the implication is that if you purchase smart management we may run stuff on your hosts using RHC regardless of whether you have that option enabled in the RHC dashboard.

aleccohan: oh!  @jharting so the reason I added that to the check is to fail if rhc is not enabled AND the customer doesn't have the smart_management entitlement, I was under the impression that if you didn't have rhcEnabled on the dashboard and didn't have the entitlement you wouldn't be able to run any playbooks  (either through RHC or Satellite)

src/remediations/controller.fifi.js
```diff
@@ -26,7 +26,18 @@ exports.connection_status = errors.async(async function (req, res) {
         throw new errors.Forbidden();
     }
 
-    const status = await fifi.getConnectionStatus(remediation, req.identity.account_number, req.entitlements.smart_management);
+    const rhcEnabled = await fifi.checkRhcEnabled();
```
jharting, line 29: might be better to do this in parallel with the db query to speed up the operation

src/remediations/__snapshots__/fifi.integration.js.snap
```diff
@@ -75,6 +75,73 @@ exports[`FiFi connection status get connection status with false smartManagement
...
+            \\"system_count\\": 1,
+            \\"connection_status\\": \\"no_executor\\"
+        },
+        {
+            \\"endpoint_id\\": null,
+            \\"executor_id\\": null,
+            \\"executor_type\\": \\"RHC\\",
+            \\"executor_name\\": null,
+            \\"system_count\\": 2,
+            \\"connection_status\\": \\"disabled\\"
```
jharting, line 139: Even though this is reported as disabled to the user, if they submit the pre-flight will remediations still attempt to execute playbook on these hosts? If not, pls point me to the code that takes care about that.

src/remediations/controller.fifi.js
jharting: Have you considered using the new DISABLED state instead? I.e. regardless of whether I have smart management or not, if I have rhc-connected hosts with RHC disabled, I see x hosts in the rhc-connected-disabled bucket in preflight. I think that ways users are more likely to find the switch they need to flip than if all they get is a 403.

src/remediations/__snapshots__/fifi.integration.js.snap
aleccohan: Hey!  so yes.  the "execute" method will only run the executors marked with 'connected'

--> right here we filter out all NON connected executors https://github.com/RedHatInsights/insights-remediations/blob/master/src/remediations/fifi.js#L140

src/remediations/controller.fifi.js
aleccohan: Will do!  making the change now.

src/remediations/__snapshots__/fifi.integration.js.snap
jharting: I see, thanks!

src/remediations/controller.fifi.js
aleccohan: I agree!  I like this alot better than sending a 403.  The newly pushed changes should remedy this!

src/remediations/controller.fifi.js
```diff
@@ -45,7 +53,18 @@ exports.executePlaybookRuns = errors.async(async function (req, res) {
...
+
+    if (!rhcEnabled && !req.entitlements.smartManagement) {
+        throw new errors.Forbidden();
+    }
```
jharting, line 60: The same changes you made to `connection_status` are also applicable here.

aleccohan: @jharting mirrored changes from connection_status in "execute" logic.

jharting (approved):

PR #40, opened 2021-05-06
[QUICK_FIX]:  Update URI generation for some of the dispatcher methods

aleccohan:
Update URI generation for some of the dispatcher methods.    
    
    
    
    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/connectors/dispatcher/impl.js
+++ b/src/connectors/dispatcher/impl.js
@@ -8,7 +8,17 @@ const {host, insecure, auth} = require('../../config').dispatcher;
 const Connector = require('../Connector');
 const metrics = require('../metrics');
 
-const QSOPTIONS = { encode: false };
+const FILTEROPTIONS = { encode: true };
+const FIELDOPTIONS = { encode: true, indices: false };
+
+function generateQueries (filter, fields) {
+    const queries = [
+        qs.stringify(filter, FILTEROPTIONS),
+        qs.stringify(fields, FIELDOPTIONS)
+    ];
+
+    return _.join(queries, '&');
+}
 
 module.exports = new class extends Connector {
     constructor () {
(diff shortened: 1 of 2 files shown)


src/connectors/dispatcher/impl.js
```diff
@@ -8,7 +8,17 @@ const {host, insecure, auth} = require('../../config').dispatcher;
 const Connector = require('../Connector');
 const metrics = require('../metrics');
 
-const QSOPTIONS = { encode: false };
+const FILTEROPTIONS = { encode: true };
+const FIELDOPTIONS = { encode: true, indices: false };
```
jharting, line 12: filter does not use indices either i.e. you can simplify by having one options constant (with indices disabled)

src/connectors/dispatcher/impl.js
```diff
@@ -8,7 +8,17 @@ const {host, insecure, auth} = require('../../config').dispatcher;
...
+
+function generateQueries (filter, fields) {
+    const queries = [
```
jharting, line 15: So we have

1. stringifying each parameter
1. construction of array of two sets of query params
1. concating the array with results

That's an OK way to do it. A bit cleaner and simpler way to do this would be to:

1. Construct and object with both filter and fields objects nested
2. Call stringify on this object.

E.g.
```
return qs.stringify({
    filter: filter.filter,
    fields: fields.fields
}, OPTIONS)
```

Of course you could even deconstruct the values in argument definitions, etc.

jharting (approved):
Looks good to me. I left a few styling suggestions.

PR #51, opened 2021-05-17
QUICKFIX: remove empty no executor state from preflight check

aleccohan:
In the event that there are no "executorless" systems when the pre-flight check is called.  Remove the empty `no-executor` status field from the returned data.
    
    
    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -242,21 +242,24 @@ function defineRHCEnabledExecutor (satellites, smart_management, rhc_enabled) {
     const satlessExecutor = _.find(satellites, satellite => satellite.id === null);
     if (satlessExecutor) {
         const partitionedSystems = sortRHCSystems(satlessExecutor, smart_management);
+        _.remove(satellites, executor => executor === satlessExecutor); // Remove redundant satless executor
 
         if (!_.isEmpty(partitionedSystems[0])) {
             satellites.push({id: null, systems: partitionedSystems[0], type: 'RHC', rhcStatus: (rhc_enabled) ? CONNECTED : DISABLED});
         }
 
-        satlessExecutor.systems = partitionedSystems[1];
-
         if (!smart_management) {
-            satlessExecutor.systems = _.filter(partitionedSystems[1], system => !_.isUndefined(system.rhc_client) && !system.marketplace);
-            const noSmartManagement = _.filter(partitionedSystems[1], system => _.isUndefined(system.rhc_client) && !system.marketplace);
+            const noSmartManagement = _.filter(partitionedSystems[1], system => !system.marketplace);
             if (!_.isEmpty(noSmartManagement)) {
                 satellites.push({id: null, systems: noSmartManagement, type: 'RHC', rhcStatus: 'no_smart_management'});
             }
 
             const rhcNotConfigured = _.filter(partitionedSystems[1], system => _.isUndefined(system.rhc_client) && system.marketplace);
+            if (!_.isEmpty(rhcNotConfigured)) {
+                satellites.push({id: null, systems: rhcNotConfigured, type: 'RHC', rhcStatus: 'no_rhc'});
+            }
+        } else {
+            const rhcNotConfigured = _.filter(partitionedSystems[1], system => _.isUndefined(system.rhc_client));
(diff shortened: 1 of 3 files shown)


src/remediations/fifi.js
```diff
@@ -247,19 +247,22 @@ function defineRHCEnabledExecutor (satellites, smart_management, rhc_enabled) {
             satellites.push({id: null, systems: partitionedSystems[0], type: 'RHC', rhcStatus: (rhc_enabled) ? CONNECTED : DISABLED});
         }
 
-        satlessExecutor.systems = partitionedSystems[1];
+        satlessExecutor.systems = _.filter(partitionedSystems[1], system => _.isUndefined(system.rhc_client) && !system.marketplace);
+        if (_.isEmpty(satlessExecutor.systems)) {
+            // If there are no "satless" executors remove this executor from list
+            satellites = _.difference(satellites, satlessExecutor);
```
jharting, line 253: Does this actually work? The docs for `_.difference` says it should be used on two arrays.

src/remediations/fifi.js
```diff
@@ -247,19 +247,22 @@ function defineRHCEnabledExecutor (satellites, smart_management, rhc_enabled) {
             satellites.push({id: null, systems: partitionedSystems[0], type: 'RHC', rhcStatus: (rhc_enabled) ? CONNECTED : DISABLED});
         }
 
-        satlessExecutor.systems = partitionedSystems[1];
+        satlessExecutor.systems = _.filter(partitionedSystems[1], system => _.isUndefined(system.rhc_client) && !system.marketplace);
```
jharting, line 250: Is the satless executor redundant now that we have `no_rhc` and `no_smart_management` states? The 2 of themshould capture all subtypes of "direct connection" hosts, or?

aleccohan: @jharting so I "think" that the satless executor state could still be a valid state for the pre-flight check since it is technically possible to have a system that both.
1. does not have a rhc_client_id and has not been registered with cloud connector.
2. is not a marketplace system and the Customer does not have the smart_management entitlement

This would definitely be a rare occurrence though.

jharting: So if a system satisfies both `no_rhc` and `no_smart_management` at the same time then it does not end up in either of those two buckets but instead goes to a third `satless executor` one?

Also, from reading the code it looks like if a customer has smart management, the host is not a marketplace one and does not have `rhc_client_id` then the host will end up in `satless executor` also (instead of no_rhc)

aleccohan: Yes, I believe that's right, should a system satisfy both of those requirements we have the "no executor" bucket to put it in.

Also Ah!  yes you're right.  I'm going to remedy that, ty.

jharting: > Yes, I believe that's right, should a system satisfy both of those requirements we have the "no executor" bucket to put it in.

Does `no_executor` that map to any of the states in https://marvelapp.com/prototype/6cf654e/screen/77431133 ?

aleccohan: I believe it's currently mapping to it's old state, i.e. Cannot remediate because of direct connect.  I'm chatting with Robin rn to make it something else.

src/remediations/fifi.js
```diff
@@ -242,16 +242,14 @@ function defineRHCEnabledExecutor (satellites, smart_management, rhc_enabled) {
     const satlessExecutor = _.find(satellites, satellite => satellite.id === null);
     if (satlessExecutor) {
         const partitionedSystems = sortRHCSystems(satlessExecutor, smart_management);
+        _.remove(satellites, satlessExecutor); // Remove redundant satless executor
```
jharting, line 245: Based on https://lodash.com/docs/4.17.15#remove the function expects a predicate as the second parameter. Even if this works somehow it would be better to modify this so that we do not depend on undocumented code.

jharting (approved):

PR #53, opened 2021-05-17
QUICKFIX: Update playbok run status based on executors (Due to RHC changes)

aleccohan:
Update playbook run status based on executors status.  This is based on RHC changes, since we don't store the status of RHC runs in the remediations db.

Eventually I would like to replace this logic with logic in the Remediations-consumer so all updates can be done over kafka for the run status.

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/remediations/fifi.js
```diff
@@ -68,6 +69,32 @@ exports.sortSystems = function (systems, column = 'system_name', asc = true) {
     return _.orderBy(systems, column, (asc) ? 'asc' : 'desc');
 };
 
+
+// TODO: Replace this logic with logic in the remediations-consumer
+exports.updatePlaybookRunsStatus = async function (playbook_runs) {
```
jharting, line 74: this function does not look like it needs to be async

src/remediations/fifi.js
```diff
@@ -68,6 +69,32 @@ exports.sortSystems = function (systems, column = 'system_name', asc = true) {
...
+    _.forEach(playbook_runs, run => {
+        if (_.some(run.executors, executor => executor.status === FAILURE)) {
+            run.status = FAILURE;
```
jharting, line 77: This logic seems to differ from what's done in https://github.com/RedHatInsights/remediations-consumer/blob/f40a63289dedff07c0776ca03bee6d3deef8798f/src/handlers/receptor/sharedQueries.ts#L28. Is that on purpose?

aleccohan: Yes, this was on purpose.  Due to the fact that an RHC `playbook_run` will not be updated at the top level I figured I would include the logic for RUNNING and PENDING along with the 3 FINAL_STATES so that the status of the job makes more sense to the user.

aleccohan: @jharting Gotcha!  Sorry I missed that.  This most recent push should have fixed the priority.

src/remediations/fifi.js
```diff
@@ -68,6 +69,30 @@ exports.sortSystems = function (systems, column = 'system_name', asc = true) {
...
+        } else if (_.some(run.executors, executor => executor.status === CANCELED)) {
+            run.status = CANCELED;
+        } else {
+            if (_.some(run.executors, executor => executor.status === RUNNING)) {
+                run.status = RUNNING;
+            }
+
+            if (_.every(run.executors, executor => executor.status === PENDING)) {
```
jharting, line 85: what if there's 1 pending, 2 success?

src/remediations/fifi.js
```diff
@@ -68,6 +69,30 @@ exports.sortSystems = function (systems, column = 'system_name', asc = true) {
     return _.orderBy(systems, column, (asc) ? 'asc' : 'desc');
 };
 
+
+// TODO: Replace this logic with logic in the remediations-consumer
+exports.updatePlaybookRunsStatus = function (playbook_runs) {
+    _.forEach(playbook_runs, run => {
+        if (_.some(run.executors, executor => executor.status === FAILURE)) {
+            run.status = FAILURE;
+        } else if (_.some(run.executors, executor => executor.status === CANCELED)) {
```
jharting, line 78: might be cleaner to isolate the logic into its own function and use return upon finding the desired states instead of all the else clauses

PR #60, opened 2021-06-01
RHCLOUD-13852: Add additional states to pre-flight check

aleccohan:
RHCLOUD-13852: Add additional states to pre-flight check.  These states are to designate which Direct connected systems are "regular" RHC systems and which are "Marketplace" systems.

JIRA:  https://issues.redhat.com/browse/RHCLOUD-13852

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -383,13 +383,17 @@ async function fetchReceptorStatus (receptor, account) {
     return _.get(result, 'status', null);
 }
 
-function getName (executor) {
+function getName (executor, smart_management) {
     if (executor.source) {
         return executor.source.name;
     }
 
     if (executor.id) {
         return `Satellite ${executor.id}`;
+    }
+
+    if (!smart_management && !_.includes([null, 'no_smart_management'], executor.rhcStatus)) {
+        return 'Direct connection - AWS Marketplace';
     }
 
     return null;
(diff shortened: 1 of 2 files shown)


src/remediations/fifi.js
```diff
@@ -392,6 +392,10 @@ function getName (executor) {
         return `Satellite ${executor.id}`;
     }
 
+    if (!smart_management && _.includes([CONNECTED, DISABLED, 'no_rhc'], executor.rhcStatus)) {
```
Jason-RH, line 395: Is there a reason to check if these three values are included instead of just checking that executor.rhcStatus is not null?

aleccohan: @Jason-RH good question, yes, because the "no_smart_management" state can't be a marketplace system it has to be part of the check.

Though, I realize now that It would be cleaner to check against NOT being [null, no_smart_management] instead of this.  I'll make that change

Jason-RH (approved):
lgtm. I'm certainly not a lodash expert so I'm not sure if checking null that way is normal, but that's probably just a nit picky thing.

PR #143, opened 2021-10-05
Implement proper reboot support for advisories

semtexzv:

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/src/generator/__snapshots__/errata.integration.js.snap
+++ b/src/generator/__snapshots__/errata.integration.js.snap
@@ -19,7 +19,8 @@ exports[`aggregates multiple errata into a single play 1`] = `
   hosts: \\"68799a02-8be9-11e8-9eb6-529269fb1459.example.com\\"
   vars:
     insights_issues: \\"--advisory RHSA-2017:1382 --advisory RHSA-2017:1852 --advisory RHSA-2017:2679 --advisory RHSA-2018:0502\\"
-    insights_signature_exclude: \\"/hosts,/vars/insights_signature,/vars/insights_issues\\"
+    requres_reboot: \\"true\\"
+    insights_signature_exclude: \\"/hosts,/vars/insights_signature,/vars/insights_issues,/vars/requires_reboot\\"
     insights_signature: !!binary |
       TFMwdExTMUNSVWRKVGlCUVIxQWdVMGxIVGtGVVZWSkZMUzB0TFMwS1ZtVnljMmx2YmpvZ1IyNTFV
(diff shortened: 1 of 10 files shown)


src/generator/__snapshots__/errata.integration.js.snap
```diff
@@ -60,7 +61,7 @@ exports[`aggregates multiple errata into a single play 1`] = `
     - when: check_out.rc == 100
       name: set reboot fact
       set_fact:
-        insights_needs_reboot: true
+        insights_needs_reboot: \\"{{requires_reboot}}\\"
```
aleccohan, line 64: @semtexzv was this change taken as an example from another service?  From my knowledge I don't believe any other service has added a var in their templates to replace the "insights_needs_reboot" field.

src/generator/__snapshots__/patchman.integration.js.snap
```diff
@@ -59,57 +60,7 @@ exports[`patchman aggregates multiple advisories into a single play 1`] = `
...
-        - name: Wait for system to boot up
-          local_action:
-            module: wait_for
-            host: \\"{{ hostvars[inventory_hostname]['ansible_host'] | default(hostvars[inventory_hostname]['ansible_ssh_host'], true) | default(inventory_hostname, true) }}\\"
-            port: \\"{{ hostvars[inventory_hostname]['ansible_port'] | default(hostvars[inventory_hostname]['ansible_ssh_port'], true) | default('22', true) }}\\"
-            delay: 15
-            search_regex: OpenSSH
-            timeout: 300
-          become: false
+        insights_needs_reboot: \\"{{requires_reboot}}\\"
```
aleccohan, line 63: @semtexzv I don't believe this is having the desired effect.  The Dynamic Variables defined in the VARS section are filled in at playbook runtime, and therefore would skip the requires_reboot inclusion that happens in Remediations at the time of playbook generation.

aleccohan (changes requested):
@semtexzv I left a couple comments regarding the placement of the "requires_reboot" var in the play template.  I don't think this placement will work with the current logic that we implement in Remediations since that "reboot" snippet is added in at playbook generation time.  

Additionally could you include a description for this PR and it's intended goals?  Up until very recently it was my understanding that all patch issues were required reboot so this is a rather big change.

PR #152, opened 2021-10-19
SPM-1118: Remove Patch package verification over API

Josca:
- in vmaas third-party is false by default, so it failed for all epel packages

    ## Secure Coding Checklist
    - [x] Input Validation
    - [x] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/connectors/patchman/impl.js
```diff
@@ -32,6 +32,25 @@ module.exports = new class extends Connector {
         ).then(res => _.get(res, ['data'], null));
     }
 
+    getPackage (id, refresh = false) {
```
aleccohan, line 35: Because of the new endpoint would you be able to add some unit-tests in the src/connectors/patchman/impl.unit.js file?

src/issues/PackageHandler.js
```diff
@@ -3,14 +3,14 @@
...
-const vmaas = require('../connectors/vmaas');
+const patchman = require('../connectors/patchman');
 const packageResolver = new(require('../resolutions/resolvers/PackageResolver'))();
 const cveFactory = new(require('../generator/factories/CVEFactory'))();
 
 module.exports = class PackageHandler extends Handler {
 
     getIssueDetailsInternal (id) {
-        return vmaas.getPackage(id.issue);
+        return patchman.getPackage(id.issue);
```
aleccohan, line 13: the src/connectors/patchman/impl.js file no longer contains a "getPackage" function so this operation will fail every time with a "UNKNOWN ISSUE" error.

src/generator/package.integration.js
```diff
@@ -70,63 +70,3 @@ describe('patchman - package using patch-package', function () {
         }]);
     });
 });
-
```
aleccohan, line 73: I would prefer that we don't remove testing code, with the current implementation we should still be able to use these tests to check that the actual generation process worked, even if we're not reaching out to the patchman API.

aleccohan (changes requested):
Hey, just a couple of changes before I merge, the rest of the logic looks alright.  Most of the changes have to do with deprecating tests that could still be useful when testing locally

src/connectors/vmaas/mock.js
```diff
@@ -44,27 +44,6 @@ const CVES = {
 
 };
 
-const PACKAGES = {
```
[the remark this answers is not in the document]
Josca: It's better not to include code which is not really used. If we start to use api again we can simply move back and this related mocking code. Do you agree?

src/issues/PackageHandler.js
Josca: Fixed.

PR #406, opened 2024-01-08
RHINENG-5482 Fixed error in identity header handling

rexwhite:
RHINENG-5482: Fixed silly error in identity header processing.

### DIFF
--- a/src/middleware/identity/userIdentity.js
+++ b/src/middleware/identity/userIdentity.js
@@ -1,18 +1,17 @@
 'use strict';
 
 const errors = require('../../errors');
-const _ = require('lodash');
 
 module.exports = function (req, res, next) {
-    switch (req.identity.type) {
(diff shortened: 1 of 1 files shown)


src/middleware/identity/userIdentity.js
```diff
@@ -3,12 +3,21 @@
...
-    }
+    switch(req.identity.type) {
+        case 'User':
+            if (!req.identity?.user?.username || req.identity?.user?.is_internal === undefined) {
+                return next(new errors.Forbidden('Supplied identity invalid'));
+            }
+            break;
+
+        case 'ServiceAccount':
+            if (!req.identity?.service_account?.username) {
```
vkrizan, line 14: Is there a place elsewhere where the username is retrieved from the header? This function only checks the header.

PR #433, opened 2024-02-20
RHIENG-4918: Add play that generates reports for Compliance(ssg) issues

marleystipich2:
If everything looks good, the new compliance/generateReport.yml needs to be signed and then I'll update the test snapshots 

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/templates/static/compliance/generateReport.yml
```diff
@@ -0,0 +1,31 @@
+- name: generate Compliance report
+  hosts: "@@HOSTS@@"
+  become: true
+  gather_facts: false
+  vars:
+    insights_signature_exclude: /hosts,/vars/insights_signature
+    insights_signature: !!binary |
```
marleystipich2, line 7: I added a fake signature... Wasn't sure if I should just leave this field blank but figured it wouldn't matter since this will be updated when the content is signed 🤔

src/generator/generator.controller.js
```diff
@@ -188,6 +195,11 @@ function addPostRunCheckIn (plays) {
     return [...plays, new SpecialPlay('special:post-run-check-in', hosts, templates.special.postRunCheckIn)];
 }
 
+function addComplianceReportPlay (plays) {
+    const hosts = _(plays).flatMap('hosts').uniq().sort().value();
+    return [...plays, new SpecialPlay('special:generate-report', hosts, templates.compliance.generateReport)];
```
vkrizan, line 200: Name of the special play might be `compliance:generate-report`.

src/generator/generator.controller.js
```diff
@@ -188,6 +195,11 @@ function addPostRunCheckIn (plays) {
     return [...plays, new SpecialPlay('special:post-run-check-in', hosts, templates.special.postRunCheckIn)];
 }
 
+function addComplianceReportPlay (plays) {
+    const hosts = _(plays).flatMap('hosts').uniq().sort().value();
```
vkrizan, line 199: Retrieval of hosts is from the existing play here is a bit odd and done several times. This might need some refactoring ideas. What are your thoughts?

src/generator/generator.controller.js
```diff
@@ -89,6 +89,13 @@ exports.playbookPipeline = async function ({issues, auto_reboot = true}, remedia
         issues = addPostRunCheckIn(issues);
     }
 
+    // Add play that generates a new Compliance report when there are Compliance(ssg) issues  
+    const complianceIssues = _.filter(issues, issue => issue.id.app == 'ssg');
```
vkrizan, line 93: Does it need to grab all compliance issues to see whether there is a one occurrence?

src/templates/static/compliance/generateReport.yml
vkrizan: :slightly_smiling_face: rather leave it blank.

vkrizan:
Is it possible to add a unit test?

src/generator/generator.controller.js
marleystipich2: Ah yes, you're right. Thanks!

src/generator/generator.controller.js
marleystipich2: Hmmm yeah we only need to check for one occurrence. I'll update this too. Thanks for the review!

marleystipich2:
> Is it possible to add a unit test?

@vkrizan 
[This snapshot](https://github.com/RedHatInsights/insights-remediations/blob/master/src/generator/__snapshots__/ssg.integration.js.snap) will be updated with the new `generateReport` play at the end of it(once I update the snapshots) so the tests in [/src/generator/ssg.integration.js](https://github.com/RedHatInsights/insights-remediations/blob/master/src/generator/ssg.integration.js) will test to make sure that the new play is included for `ssg` issues. 

I can probably add a test that tests this directly though just to be safe.

vkrizan:
The snapshot testing should be enough. Thanks for pointing that out.

marleystipich2:
@rexwhite Can you take a look at this when you get a chance? It might be ready for you to sign but let me know if you have permission issues editing the branch. (Commenting here in the spirit of not overusing slack 😄 )

src/generator/generator.controller.js
rexwhite: Remediations in general seems really bad about making redundant calls, to inventory in particular, but I'd prefer to address that in a separate task.  Basically we need to review each endpoint and clean things up systematically.

rexwhite (approved):
Looks good!  I'll sign the special play and we can merge this once this once you update the test snapshots.

PR #442, opened 2024-03-11
RHINENG-4942: Prevent duplicate or empty remediation names

marleystipich2:
The purpose of this is to prevent duplicate remediation names within an organization. It will also prevent empty or null remediation names. The parent ticket contains more info on why this is necessary: https://issues.redhat.com/browse/RHINENG-4942

The migration will do the following: 
- Run a query that will search for duplicate remediation names or null remediations names within an **organization** and append an incrementing number to the end of each name to ensure that there are no more duplicate or null names. (Will update about 3562 names in production)

- Add a UNIQUE constraint on the `name` column on the `remediations` table 
- Add a NOT NULL constraint on the `name` column on the `remediations` table

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management


src/remediations/controller.write.js
```diff
@@ -170,7 +170,24 @@ exports.patch = errors.async(async function (req, res) {
...
+                throw new errors.BadRequest('DUPLICATE_NAME', `Remediation name "${name}" for user "${username}" has already been taken`);
+            }
+
             remediation.name = name;
+        } 
+
+        // Do you not allow users to update a remediation name to empty string or null
+        if (name === "" || name === null) {
+            throw new errors.BadRequest('INVALID_NAME', "Remediation name cannot be empty"); 
         }
```
vkrizan, line 191: What about merging these two conditions for the `name` together?

src/remediations/controller.write.js
```diff
@@ -170,7 +170,24 @@ exports.patch = errors.async(async function (req, res) {
...
 
         if (name) {
+            const existing_remediation = await db.remediation.findOne({
+                attributes: ['id'],
+                where: { created_by: username, name: name },
```
vkrizan, line 175: Isn't this missing a tenant condition?

vkrizan (changes requested):

src/remediations/controller.write.js
```diff
@@ -170,7 +170,24 @@ exports.patch = errors.async(async function (req, res) {
...
+            }, {
+                transaction
+            });
+
+            // Do not allow users to use the same name for more than one remediation
+            if (existing_remediation) {
+                throw new errors.BadRequest('DUPLICATE_NAME', `Remediation name "${name}" for user "${username}" has already been taken`);
```
vkrizan, line 182: Does it need to inform about the username? And does it even have to be tied to a specific user in an org?

src/remediations/controller.write.js
```diff
@@ -170,7 +170,24 @@ exports.patch = errors.async(async function (req, res) {
...
         if (name) {
+            const existing_remediation = await db.remediation.findOne({
+                attributes: ['id'],
+                where: { created_by: username, name: name },
+            }, {
+                transaction
+            });
+
+            // Do not allow users to use the same name for more than one remediation
+            if (existing_remediation) {
```
vkrizan, line 181: What is the motivation to prevent duplicate names of playbooks? The ticket does not include any details on this.

marleystipich2: Here's the main ticket for this: https://issues.redhat.com/browse/RHINENG-4942

The 2 subtasks don't have much info but this explains it better.

src/remediations/controller.write.js
vkrizan: account / org check

src/remediations/controller.write.js
marleystipich2: My understanding was that we don't want to allow remediations to have the same name "on an account basis" so basically a user shouldn't be able to create a remediation or update a remediation using the same name as an existing one under their account.

src/remediations/controller.write.js
marleystipich2: Ah yes, an org id check would make sense here. I'll add that.

marleystipich2:
Converted to draft because I'm going to combine this with the changes for the POST operation too: https://issues.redhat.com/browse/RHINENG-8655

There's a lot of overlapping code so I think it makes more sense.

src/remediations/controller.write.js
rexwhite: `if (!_.is_empty(name))` would be more consistent with the rest of the code in this file.

marleystipich2: Yep, good idea I'll use this! Do you mean `_.isEmpty()` though?

rexwhite: Whoops!   Yeah, `_.isEmpty()` 😬

src/remediations/controller.write.js
```diff
@@ -55,6 +55,27 @@ async function validateNewActions(add) {
     });
 }
 
+async function validateRemediationName(name, tenant_org_id, username, transaction) {
```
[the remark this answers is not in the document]
marleystipich2: I did consider the UNIQUE constraint but figured this was more consistent with the rest of the code. 

@vkrizan That's a good point actually.. Would the db constraint cause issues with existing data? And if we keep this current implementation, I don't think it would catch any existing duplicate names unless the remediation is updated at some point. Not sure if this is a good thing or not?

src/remediations/controller.write.js
marleystipich2: Hmm good point.. Yeah, I'm free to talk about this today or we can do tomorrow.

src/remediations/controller.write.js
```diff
@@ -113,6 +132,12 @@ exports.create = errors.async(async function (req, res) {
     const id = uuid.v4();
 
     const result = await db.s.transaction(async transaction => {
+        if ( !_.isUndefined(name) ) {
+            await validateRemediationName(name, req.user.tenant_org_id, req.user.username, transaction);
+        } else {
+            throw new errors.BadRequest('UNDEFINED_NAME', 'Remediation name must be defined upon creation');
+        }
+
```
marleystipich2, line 140: When creating a new remediation, I think it will be necessary to throw this error when the name is undefined because this would result in a `null` name which we don't want.

rexwhite:
Consider removing this as well: src/remediations/controller.read.js:131

marleystipich2:
Still in progress but just wanted to push what I have so far. Need to test the migration still and make sure the queries in the migration do what they're supposed to do.

src/remediations/write.integration.js
```diff
@@ -149,7 +149,7 @@ describe('remediations', function () {
         });
 
         test('creates a new remediation with 20k systems', async () => {
-            const name = 'new remediation with issues';
+            const name = `new remediation with issues ${Math.random()}`;
```
vkrizan, line 152: Can this conflict at some point in future test runs, or, is it cleaned up?

db/migrations/20240318200010-update-exisiting-remediation-names.js
```diff
@@ -0,0 +1,54 @@
...
+        try {
+            //===================================================
+            // The next 2 queries will update existing remediation names to ensure that there are
+            // no duplicate remediation names within an organization
+            //===================================================
+
+            // Find duplicate remediation names within an organization and add the username to the beginning of the name
+            // Example: 'test1@redhat.com: duplicate name', 'test2@redhat.com: duplicate name', 'test3@redhat.com: duplicate name'...
+            const update_dups_in_org = `UPDATE remediations
+                                        SET name = created_by || ': ' || name
```
vkrizan, line 15: Why do we append username here?
Can this rather prepend creation/modification time in the unix format? That could resolve the duplicates within an account and would not require names with incrementing numbers.

db/migrations/20240318200010-update-exisiting-remediation-names.js
```diff
@@ -0,0 +1,54 @@
...
+                                        (
+                                            SELECT name, tenant_org_id
+                                            FROM remediations
+                                            GROUP BY name, tenant_org_id
+                                            HAVING ( COUNT( name ) > 1 )
+                                        )`
+
+            // Find duplicate remediation names within an account and update names to include incrementing numbers at the end of the names
+            // Example: 'duplicate name1', 'duplicate name2', 'duplicate name3'...
+            const update_dups_in_account = `WITH updated_names AS
```
vkrizan, line 26: This one seems overly complicated. See my comment above.

db/migrations/20240318200010-update-exisiting-remediation-names.js
vkrizan: This can be pulled from `created_at`. Just check it's type. Looking at some migrations this might be just a date (and not datetime).

marleystipich2: The `created_at` type is `DATE` but it looks like the format is `2024-03-04T16:12:51.012Z` so I think this would work. I think one query like this would be enough to update all existing duplicate names: 
```
UPDATE remediations
SET name = created_at || ': ' || name 
WHERE (name, tenant_org_id) IN
(
           SELECT name, tenant_org_id
           FROM remediations
           GROUP BY name, tenant_org_id
           HAVING ( COUNT( name ) > 1 )
)
```

rexwhite:
The sequelize model object in src/remediations/models/remediation.js needs to be updated too.

db/migrations/20240410141239-add-constraints-on-remediation-name.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+module.exports = {
+    up: async (q, {STRING}) => {
+        const transaction = await q.sequelize.transaction()
+        try {
+            // Add unique constraint to remediations name column based on tenant_org_id
+            await q.addConstraint('remediations', {
```
rexwhite, line 8: This will fail if someone adds a record with a null or duplicate name between the time the previous migration ran and this migration runs - that's a pretty narrow window but it might be worth thinking about.  BTW I'm not sure the exact SQL this will generate so this might not be an issue, or we might need to include another run of what you did in the previous migration as part of this transaction.

src/remediations/write.integration.js
rexwhite: The test db is rebuilt at the beginning of the test run, so this was actually fine the way it was.

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+            .set(auth.testWrite)
+            .send({name})
+            .expect(201);
+
+            const {body} = await request
+            .post('/v1/remediations')
+            .set(header)
+            .set(auth.testWrite)
+            .send({name})
+            .expect(500);
```
rexwhite, line 254: 500 is an internal server error.  I know the remediations code is a bit fragile but we should probably be returning a 400 (with an appropriate message) for duplicate remediation plan name, right?  (500 would be a bug...)

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
                 title: 'Unsupported Content-Type application/xml (location: undefined, path: undefined)'
             }]);
         });
+
+        test('500s when remediation is created with name that is used by existing remediation created by the same user', async () => {
```
rexwhite, line 238: 400?

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+
+            const {body} = await request
+            .post('/v1/remediations')
+            .set(header)
+            .set(auth.testWrite)
+            .send({name})
+            .expect(500);
+        });
+
+        test('500s when remediation is created with name used by existing remediation created by a different user within the same org', async () => {
```
rexwhite, line 257: 400...

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+            .set(auth.testWrite2)
+            .send({name})
+            .expect(201);
+
+            const r2 = await request
+            .post('/v1/remediations')
+            .set(header)
+            .set(auth.testWrite)
+            .send({name})
+            .expect(500);
```
rexwhite, line 273: Same as above: we should be returning a 400 with an appropriate message.

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+        });
+
+        test('200s when remediation is created with name used by existing remediation created by a different user in a different org', async () => {
```
rexwhite, line 276: 201's...

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+
+            body.errors.should.eql([{
+                id,
+                status: 400,
+                code: 'type.openapi.validation',
+                title: 'should be string (location: body, path: name)'
+            }]);
+        });
+
+        test('500s when remediation name is undefined upon creation', async () => {
```
rexwhite, line 333: 400...

src/remediations/write.integration.js
```diff
@@ -234,6 +234,112 @@ describe('remediations', function () {
...
+            const {id, header} = reqId();
+
+            const {body} = await request
+            .post('/v1/remediations')
+            .set(header)
+            .send({auto_reboot: false})
+            .set(auth.testWrite)
+            .expect(500);
```
rexwhite, line 341: Again - if this causes "internal server error"'s then we need to fix the code.  This should be a 400 with an appropriate message.

src/remediations/write.integration.js
```diff
@@ -481,6 +587,119 @@ describe('remediations', function () {
                         title: 'At least one of "add", "name", "auto_reboot", "archived" needs to be specified'
                     }]);
                 });
+
+                test('500s when remediation is renamed to name that is used by existing remediation created by the same user', async () => {
```
rexwhite, line 591: 400...

src/remediations/write.integration.js
```diff
@@ -481,6 +587,119 @@ describe('remediations', function () {
...
+                    .send({name})
+
+                    const url = '/v1/remediations/8b427145-ac9f-4727-9543-76eb140222cd';
+
+                    const {body} = await request
+                    .patch(url)
+                    .set(header)
+                    .send({name})
+                    .set(auth.testWrite)
+                    .expect(500);
```
rexwhite, line 607: 400...

src/remediations/write.integration.js
```diff
@@ -481,6 +587,119 @@ describe('remediations', function () {
...
+                });
+
+                test('500s when remediation is renamed to a name used by existing remediation created by a different user within the same org', async () => {
```
rexwhite, line 610: 400...

src/remediations/write.integration.js
```diff
@@ -481,6 +587,119 @@ describe('remediations', function () {
...
+                    .send({name})
+
+                    const url = '/v1/remediations/8b427145-ac9f-4727-9543-76eb140222cd';
+
+                    const {body} = await request
+                    .patch(url)
+                    .set(header)
+                    .send({name})
+                    .set(auth.testWrite)
+                    .expect(500);
```
rexwhite, line 626: 400...

rexwhite (changes requested):

db/migrations/20240410141239-add-constraints-on-remediation-name.js
```diff
@@ -0,0 +1,35 @@
...
+                name: 'name_and_tenant_org_id',
+                type: 'UNIQUE',
+                fields: ['name', 'tenant_org_id']
+            }, { transaction });
+
+            // Add NOT NULL constraint to name
+            // Fixing duplicates should make sure there are no NULL name values because either
+            await q.changeColumn('remediations', 'name', {type: STRING, allowNull: false}, { transaction });
```
rexwhite, line 16: Unless I missed it we never changed the existing records with null names...

rexwhite (commented):
The sequelize model object in src/remediations/models/remediation.js needs to be updated too.

db/migrations/20240318200010-update-exisiting-remediation-names.js
```diff
@@ -0,0 +1,34 @@
...
+            // no duplicate remediation names within an organization
+            //===================================================
+
+            // Find duplicate remediation names within an organization and add the username to the beginning of the name
+            // Example: '2024-01-02T06:06:55.759Z: duplicate name', '2024-02-19T16:23:47.267Z: duplicate name', '2023-11-27T16:29:40.969Z: duplicate name'...
+            const update_dups_in_org = `UPDATE remediations
+                                        SET name = created_at || ': ' || name 
+                                        WHERE (name, tenant_org_id) IN
+                                        (
+                                            SELECT name, tenant_org_id
```
rexwhite, line 18: I think:
```
SELECT
  coalesce(name, 'Unnamed Remediation Plan') AS name,
  tenant_ord_id
...
```
would also include the entries where `name` is NULL if you do something like:
```
HAVING COUNT(name) > 1 OR name IS NULL
```

db/migrations/20240318200010-update-exisiting-remediation-names.js
vkrizan: Yes, looks like it should be a datetime, as per [docs](https://sequelize.org/api/v6/class/src/data-types.js~date). I'd transform that to unix time to get a number, and add it to the end (not to the start of the original name). @rexwhite thought?

db/migrations/20240318200010-update-exisiting-remediation-names.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+module.exports = {
+    up: async (q, {STRING}) => {
+        const transaction = await q.sequelize.transaction()
```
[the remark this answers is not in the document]
vkrizan: I'd suggest to do an analysis to know what volume we are dealing with. @marleystipich2 or have you already done it? Can you please send out the results privately? Thank you.

src/remediations/write.integration.js
marleystipich2: Yeah, I was wondering about this. Maybe the constraint isn't being setup correctly because a 500 error(not 400) is being returned when you try to create/update a remediation with a duplicate name. I'm gonna look into the constraint more.

src/remediations/write.integration.js
marleystipich2: This test was failing and I thought it was because it was creating more than one remediation using the same name in this test but it was actually failing because a previous test in this same file already uses this name: `test('creates a new remediation with issues'`. 
So it thinks the name `new remediation with issues` is a duplicate but if I change it to just `new remediation with issues 2` it passes.

db/migrations/20240410141239-add-constraints-on-remediation-name.js
marleystipich2: I was for some reason thinking that this would be resolved by updating the duplicates but I guess there might be a null name that isn't a duplicate within an org 🤔 I can add a query that updates remaining null values! I'm guessing there won't be a lot so it probably wouldn't add too much time.

db/migrations/20240318200010-update-exisiting-remediation-names.js
marleystipich2: I think this would work too or even this:
```
UPDATE remediations
SET name = created_at || ': ' || name 
WHERE (name, tenant_org_id) IN
(
           SELECT name, tenant_org_id
           FROM remediations
           GROUP BY name, tenant_org_id
           HAVING ( COUNT( name ) > 1 )
)
OR name IS NULL
```

db/migrations/20240318200010-update-exisiting-remediation-names.js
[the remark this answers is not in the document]
rexwhite: I'd suggest we not implement the migration in a way that's susceptible to stuck deployments.  There are enough records in need of updating that this is a legitimate concern.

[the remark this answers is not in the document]
marleystipich2: Agreed. I'll move some things around snd make the changes from the other comments and see what you guys think.

db/migrations/20240318200010-update-exisiting-remediation-names.js
rexwhite: Here's my latest absurd suggestion:
```
WITH dups AS (
    SELECT
        a.id,
        a.name,
        a.tenant_org_id,
        ROW_NUMBER() OVER (PARTITION BY a.tenant_org_id, a.name) AS number
    FROM remediations a
    JOIN (
        SELECT name, tenant_org_id
        FROM remediations
        GROUP BY name, tenant_org_id
        HAVING COUNT(*) > 1 OR name IS NULL
        LIMIT 1000
    ) b
    ON (a.name = b.name OR (a.name IS NULL AND b.name IS NULL)) AND a.tenant_org_id = b.tenant_org_id
)
UPDATE remediations
    SET name = coalesce(dups.name, 'Unnamed Remediation Plan') || ' - ' || dups.number
FROM dups
WHERE dups.id = remediations.id;
```
We could omit the `LIMIT` for now and see if it works in one pass.  If not, we can add it and then just repeat this until no more rows are changed.

This also goes back to using incrementing numbers instead of awkward timestamps.

src/test/index.js
```diff
@@ -62,6 +62,7 @@ exports.auth = Object.freeze({
         return id;
     }),
     testWrite: createHeader(USERS.testWriteUser.username, USERS.testWriteUser.account_number, USERS.testWriteUser.tenant_org_id, false),
+    testWrite2: createHeader('testWriteUser2', 'testWrite2', USERS.testWriteUser.tenant_org_id, false),
```
rexwhite, line 65: We should probably define this testWrite2 user in `src/connectors/users/mock.js` so it's easier to keep track of the overall test configuration.

marleystipich2:
Tests are passing besides 2 tests that are failing intermittently. Looking into this now.

rexwhite (changes requested):
There are some additional changes we should consider at some point.  For instance:

- src/remediations/controller.read.js:131
- src/remediations/remediations.queries.js:135
- src/remediations/models.js:3

There are likely other bits of code that handle null remediation plan names but I think we should handle those changes in a different PR / jira issue.

rexwhite (approved):

PR #458, opened 2024-04-01
RHINENG-9095: Added robust tests for connection status collation.

rexwhite:
There was an unhandled edge case in the code that reassembles chunked connection status requests.  This PR fixes that problem and improved tests to cover this condition.


rexwhite:
This is mostly updates to tests and a relatively minor code change.  I'm just going to merge it so QA can verify the fix...

PR #459, opened 2024-04-02
RHINENG-4860: POST remediation/:id/playbook endpoint to support many systems

rexwhite:
Added a new POST /remediations/:id/playbook endpoint to allow for large lists of system inventory ids, since otherwise the url max length is exceeded.


vkrizan (approved):
Lovely

PR #471, opened 2024-04-18
RHINENG-9557: Avoid creating a user in our image (use HOME envar instead)

rexwhite:
Upgrading node and npm necessitated creating a user with a home directory (for npm to function).  Just setting the HOME envar is sufficient, however, so let's do that instead.


marleystipich2 (approved):
LGTM 👍

PR #472, opened 2024-04-23
RHINENG-9584: Fix connection status aggregation

rexwhite:
Fixed the connection_status to correctly report system counts by connect status type for direct-connected systems.

### DIFF
--- a/src/remediations/fifi.integration.js
+++ b/src/remediations/fifi.integration.js
@@ -153,10 +153,10 @@ describe('FiFi', function () {
             // use impl version
             base.getSandbox().stub(dispatcher, 'getConnectionStatus').callsFake(dispatcher_impl.getConnectionStatus);
 
-            // replace doHttp() with our own function...
-            const spy = base.getSandbox().stub(Connector.prototype, 'doHttp');
+            // replace request() with our own function...
+            const http_request = require('../util/request');
             const dispatcherMock = require('../connectors/dispatcher/serviceMock');
-            spy.callsFake(dispatcherMock);
+            base.getSandbox().stub(http_request, 'run').callsFake(dispatcherMock);
 
 
             const result = await request
--- a/src/remediations/fifi.integration.js
+++ b/src/remediations/fifi.integration.js
@@ -163,7 +163,7 @@ describe('FiFi', function () {
             .get('/v1/remediations/dd6a0b1b-5331-4e7b-92ec-9a01806fb181/connection_status')
             .set(auth.fifi);
 
-            expect(result.body).toMatchObject({});
+            expect(result.body).toMatchSnapshot();
         });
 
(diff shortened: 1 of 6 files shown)


src/remediations/fifi.integration.js
```diff
@@ -153,10 +153,10 @@ describe('FiFi', function () {
             // use impl version
             base.getSandbox().stub(dispatcher, 'getConnectionStatus').callsFake(dispatcher_impl.getConnectionStatus);
 
-            // replace doHttp() with our own function...
-            const spy = base.getSandbox().stub(Connector.prototype, 'doHttp');
+            // replace request() with our own function...
+            const http_request = require('../util/request');
```
vkrizan, line 157: Shouldn't requirements be at the top of the module? Or is this intentionally here?

src/remediations/remediations.format_2.js
```diff
@@ -52,52 +37,79 @@ exports.connectionStatus = function (recipients) {
...
-                        system_count: recipient.systems.length,
-                        connection_status: recipientStatusMap[recipient.status]
-                    };
-            }
-        })
-        .sortBy('executor_name')
-        .value();
+    const data = _(recipients)
+    .groupBy(recipient => {
+        // key will either be one of recipientStatusMap or a recipient_id UUID
```
vkrizan, line 43: There is a combination of two unrelated things when doing the grouping. I'd suggest to filter or group by type, and then operate on each of the type separately.

src/remediations/remediations.format_2.js
```diff
@@ -52,52 +37,79 @@ exports.connectionStatus = function (recipients) {
...
+
+            case PD_TYPE_SATELLITE:
+                return `${recipient.sat_id} ${recipient.sat_org_id}`;
+
+            default:
+                return PD_STATUS_NOT_CONFIGURED;
+        }
+    })
+    .map((items, group) => {
+        const system_ids = _(items).map('systems').flatten().value();
```
vkrizan, line 56: Is this supposed to aggregate recipients of type direct by status?

rexwhite: It's a slight semantic change, but not much of a departure from how it has functioned in the past.  This endpoint has evolved somewhat to accommodate first RHC-direct systems and then RHC-satellite systems.  It's undergone greater _syntactic_ changes in the past without creating a new endpoint version so we're pretty comfortable with this.  Also, it now actually provides useful information for direct systems.

vkrizan (approved):

src/remediations/remediations.format_2.js
rexwhite: I'll restructure this a bit.  It seems to be slightly faster to handle these two cases together, as part of one stream, than to split them out and then rejoin them.  I also find the code easier to follow this way but that's just a personal preference :-)

PR #483, opened 2024-05-01
Update add noop to prod build.

shannon-donahue:
This PR moves noop module to the prod image files so that it is included in the prod build along with the test build. This is to avoid the following error when deploying when noop is required:

```
Error: Cannot find module './noop'
```

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -13,6 +13,7 @@ const prod_image_files = [
     'src/**/*.js',
     'src/**/*.yaml',
     'src/**/*.yml',
+    'src/**/noop.js',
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -13,14 +13,14 @@ const prod_image_files = [
     'src/**/*.js',
     'src/**/*.yaml',
     'src/**/*.yml',
+    'src/**/noop.js',
```
rexwhite, line 16: Ah... I wondered if anyone would ever come looking for noop.js, lol...

rexwhite (approved):

PR #485, opened 2024-05-02
Relax openapi remediation plan name restrictions pending db updates

rexwhite:
Relax the leading/trailing whitespace restrictions until we can fix the handful of database entries with "invalid" names.


marleystipich2 (approved):
LGTM

PR #486, opened 2024-05-03
RHINENG-9857: restoring regex to disallow leading/trailing whitespace

rexwhite:
Just restoring the previous regex...


marleystipich2 (approved):
LGTM!

PR #489, opened 2024-05-14
RHINENG-10058: Adding migration to remove leading / trailing tabs characters from plan names.

rexwhite:
There were a few plans with leading or trailing tab characters in their names that we missed in the previous migration.  This migration will clean those up.

### DIFF
--- a/db/migrations/20240514205117-trim-tabs-from-plan-names.js
+++ b/db/migrations/20240514205117-trim-tabs-from-plan-names.js
@@ -1,0 +1,39 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+    async up(q, Sequelize) {
+        // Fix plan names that would be duplicates once TRIM()-ed
+        const remove_dup_tabs = `
+            WITH dups AS (
+                SELECT b.name, b.id
+                FROM remediations a
+                JOIN (
+                    SELECT TRIM('\t ' FROM name), id, tenant_org_id
+                    FROM remediations
+                    WHERE name LIKE '\t%' OR name LIKE '%\t'
+                ) b
+                ON a.name = b.name AND a.tenant_org_id = b.tenant_org_id
+            )
+            UPDATE remediations
+            SET name = dups.name || ' - 1'
+            FROM dups
+            WHERE dups.id = remediations.id;
+        `;
+
+        await q.sequelize.query(remove_dup_tabs);
+
+        // trim remaining plan names
+        const remove_tabs = `
+            UPDATE remediations
+            SET name = TRIM('\t ', FROM name)
+            WHERE name LIKE '\t%' OR name LIKE '%\t';
+        `;
+
+        await q.sequelize.query(remove_tabs);
+    },
+
+    async down(q, Sequelize) {
+        // There is no turning back...
+    }
+};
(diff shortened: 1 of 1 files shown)


db/migrations/20240514205117-trim-tabs-from-plan-names.js
```diff
@@ -0,0 +1,39 @@
...
+            UPDATE remediations
+            SET name = TRIM('\t ', FROM name)
+            WHERE name LIKE '\t%' OR name LIKE '%\t';
+        `;
+
+        await q.sequelize.query(remove_tabs);
+    },
+
+    async down(q, Sequelize) {
+        // There is no turning back...
```
marleystipich2, line 37: 😂 😂

db/migrations/20240514205117-trim-tabs-from-plan-names.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+exports.up = async (q) => {
+    // manually update the following non-compliant remediation plan names:
+    const updates = {
+        "133bd18f-7134-4a83-8034-77ba15ca8327": "A newer version of kernel is installed: but not used. Bug fixes issued by this new kernel version are not applied",
```
vkrizan, line 6: Instead of writing down concrete ids from certain environment(s), I'd suggest to do the change programmatically to handle all environments consistently. That would also avoid leaking data into application's git repository.

rexwhite: While I agree, sequelize seems to want to sanitize parts of the SQL and won't let me pass a tab character to the TRIM() function.  I've already spent 6 hours trying to debug sequelize migration processing and just need to get this done since it's leading to pod restarts in prod.  If I had access to a db console I would just fix these by hand but that's not allowed, so here we are.

marleystipich2 (approved):
If we're good with this solution, I verified that the names will be updated correctly so this looks good to me 👍

db/migrations/20240514205117-trim-tabs-from-plan-names.js
rexwhite: Yeah, 🤮...  After _much_ hacking I found a gross workaround, but it's better than targeting specific ids.

db/migrations/20240514205117-trim-tabs-from-plan-names.js
```diff
@@ -0,0 +1,47 @@
...
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+    async up(q, Sequelize) {
+        // Fix plan names that would be duplicates once TRIM()-ed
+        const remove_dup_tabs = `
+            WITH dups AS (  
+                SELECT a.name, b.id  
+                FROM remediations a  
+                JOIN (  
+                    SELECT TRIM(TRIM('\\\\t' FROM name)) AS trimmed_name, id, tenant_org_id
```
vkrizan, line 12: Just wondering whether a string replace would together with a trim could be a better choice, to cover cases where tabs are in-between.

vkrizan (approved):

marleystipich2 (approved):

db/migrations/20240514205117-trim-tabs-from-plan-names.js
rexwhite: Tabs in the middle are actually allowed, as are lots of other characters - I think the idea was to allow for other codesets.

PR #500, opened 2024-06-18
Reorder anyOf statements

Ichimonji10:
GET /remediations returns one of two objects:

* RemediationList
* RemediationNameList

The openapi-generator-cli v7.6.0 python binding generator generates a distinct schema for each of these objects. When decoding a response, the bindings will attempt to match the response to each of these schemas, in the same order as given in the spec file.

RemediationNameList is a strict subset of RemediationList. If the response is matched against these schemas in this order, then the first comparison will always succeed, and the first type of object will always be returned. This is problematic.

Fix: https://issues.redhat.com/browse/RHINENG-10796

### DIFF
--- a/src/api/openapi.yaml
+++ b/src/api/openapi.yaml
@@ -47,8 +47,8 @@ paths:
             application/json:
               schema:
                 anyOf:
-                  - $ref: '#/components/schemas/RemediationNameList'
                   - $ref: '#/components/schemas/RemediationList'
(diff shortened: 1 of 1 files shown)


marleystipich2 (approved):
This change should be okay! I just wanted to make sure that this wouldn't mess with the OpenAPI validations but I don't think this will impact anything.

PR #503, opened 2024-07-01
RHINENG-11011: Fix null dereference errors

rexwhite:
Fixed the code to handle case where sequelize returns a null.

### DIFF
--- a/src/remediations/controller.read.js
+++ b/src/remediations/controller.read.js
@@ -204,7 +204,7 @@ exports.list = errors.async(async function (req, res) {
                 req.user.username
             );
 
-            playbook_runs = playbook_runs?.toJSON();
+            playbook_runs = playbook_runs?.toJSON() || {};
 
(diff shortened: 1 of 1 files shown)


marleystipich2 (approved):

src/remediations/controller.read.js
```diff
@@ -204,7 +204,7 @@ exports.list = errors.async(async function (req, res) {
                 req.user.username
             );
 
-            playbook_runs = playbook_runs?.toJSON();
+            playbook_runs = playbook_runs?.toJSON() || {};
```
vkrizan, line 207: If there are no playbook runs, don't we want to return right away?

rexwhite: Well... sort of.  This code is really bad about side effects so you have to be really careful, but we can at least break out of this 'if' block.  I was in the process of changing that when I got pulled into a series of meetings but I'll push an update here shortly...

PR #519, opened 2024-07-25
RHINENG-11246: Fixed caching to support non-JSON data

rexwhite:
Extended the REDIS caching code to support transformed non-JSON data (compliance-ssg returns raw yaml).

### DIFF
--- a/src/connectors/http.js
+++ b/src/connectors/http.js
@@ -100,19 +100,23 @@ async function run (options, useCache = false, metrics = false, responseTransfor
     }
 
     if (res.statusCode === 304) {
+        // the etag still matches so just re-insert the old value...
         log.trace({key}, 'revalidated');
         saveCachedEntry(cache.get(), key, cached.etag, cached.body);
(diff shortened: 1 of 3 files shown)


src/connectors/http.js
```diff
@@ -100,19 +100,23 @@ async function run (options, useCache = false, metrics = false, responseTransfor
...
+        return responseTransformer(res);
     }
 
+    // update cache with transformed result
     log.trace({key}, 'saved to cache');
-    saveCachedEntry(cache.get(), key, res.headers.etag, res.body);
-    return responseTransformer(res);
+    const value = responseTransformer(res);
+    saveCachedEntry(cache.get(), key, res.headers.etag, value);
+    return value;
```
marleystipich2, line 119: So because we're passing the transformed response to `saveCachedEntry`, this allows it to handle non-JSON data?

marleystipich2 (approved):
Looks great!

PR #526, opened 2024-09-03
RHINENG-12281: Removed resolved issue count subquery from playbook download code

rexwhite:
The remediation plan query used by the playbook download endpoint included a sub-query to count resolved issues, which is rather inefficient.  Since this value is never used, a parameter was added to remove it from the query for this code path.

### DIFF
--- a/src/remediations/controller.read.js
+++ b/src/remediations/controller.read.js
@@ -301,8 +301,11 @@ exports.playbook = errors.async(async function (req, res) {
     const tenant_org_id = req.identity.org_id;
     const creator = cert_auth ? null : req.user.username;
 
+    const USE_CACHE = true;
+    const EXCLUDE_RESOLVED_COUNT = false;
+
     trace.event('Get remediation plan from db (w/caching)');
-    const remediation = await queries.get(id, tenant_org_id, creator, true);
+    const remediation = await queries.get(id, tenant_org_id, creator, EXCLUDE_RESOLVED_COUNT, USE_CACHE);
 
     if (!remediation) {
         return notFound(res);
--- a/src/remediations/remediations.queries.js
+++ b/src/remediations/remediations.queries.js
@@ -237,15 +237,14 @@ exports.getPlanNames = function (tenant_org_id) {
     ]
   }
 */
-exports.get = async function (id, tenant_org_id, created_by = null, useCache = false) {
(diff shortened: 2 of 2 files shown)


src/remediations/controller.read.js
```diff
@@ -301,8 +301,11 @@ exports.playbook = errors.async(async function (req, res) {
     const tenant_org_id = req.identity.org_id;
     const creator = cert_auth ? null : req.user.username;
 
+    const USE_CACHE = true;
+    const EXCLUDE_RESOLVED_COUNT = false;
+
     trace.event('Get remediation plan from db (w/caching)');
-    const remediation = await queries.get(id, tenant_org_id, creator, true);
+    const remediation = await queries.get(id, tenant_org_id, creator, EXCLUDE_RESOLVED_COUNT, USE_CACHE);
```
vkrizan, line 308: I recommend to pass the argument values by name, to know which boolean is used for.

src/remediations/controller.read.js
```diff
@@ -301,8 +301,11 @@ exports.playbook = errors.async(async function (req, res) {
     const tenant_org_id = req.identity.org_id;
     const creator = cert_auth ? null : req.user.username;
 
+    const USE_CACHE = true;
+    const EXCLUDE_RESOLVED_COUNT = false;
```
vkrizan, line 305: Why uppercase?

vkrizan (approved):

src/remediations/controller.read.js
rexwhite: Convention is to use uppercase for things that are static / hard-coded.

src/remediations/controller.read.js
rexwhite: Javascript doesn't really do named arguments like python.  I could pass in an options object and destructure it but that just seems like overkill, plus it's extra work to support default values - didn't seem worth it 🤷🏻‍♂️

marleystipich2 (approved):
LGTM!

src/remediations/controller.read.js
vkrizan: In that case, those should be moved to the top of the file. `const` keywords denotes a constant already.

src/remediations/controller.read.js
vkrizan: Good to know.

PR #541, opened 2024-09-18
chore(clowdapp): add mosquitto MQTT to opt deps

vkrizan:

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/deployment/clowdapp.yml
+++ b/deployment/clowdapp.yml
@@ -18,6 +18,8 @@ objects:
     - compliance-ssg
     - config-manager
     - cloud-connector
+    # add MQTT mock (mosquitto) from cloud-connector
+    - mosquitto
     - patchman
     - playbook-dispatcher
     - sources-api


rexwhite (approved):
'Mosquitto' 😂...

deployment/clowdapp.yml
```diff
@@ -18,6 +18,8 @@ objects:
     - compliance-ssg
     - config-manager
     - cloud-connector
+    # add MQTT mock (mosquitto) from cloud-connector
+    - mosquitto
```
vkrizan, line 22: We need to check whether this wouldn't cause issues on stage/prod. Event cloud-connector does not lists this as an (optional) dependency, as it is used only for testing on ephemeral envs.

We can follow the same, but then the cloud-connector needs to be deployed explicitly, as `mosquitto` is pulled out of possible components of the "cloud-connector" app (as per app-interface definitions).

Note, that these lists components and not app names.

vkrizan:
@rexwhite now I've got the pun :joy:

deployment/clowdapp.yml
rexwhite: Yes - let's not break prod, please 😬...

PR #570, opened 2024-11-21
RHINENG-14308: Upgrade rpm packages

marleystipich2:

    ## Secure Coding Checklist
    - [ ] Input Validation
    - [ ] Output Encoding
    - [ ] Session Management
    - [ ] Access Control
    - [ ] Cryptographic Practices
    - [ ] Data Protection
    - [ ] Communication Security
    - [ ] System Configuration
    - [ ] Database Security
    - [ ] File Management
    - [ ] Memory Management

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,6 +4,7 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1130 AS base
 
 RUN microdnf module enable nodejs:20 && \
     microdnf install -y shadow-utils jq nodejs --nodocs && \
+    microdnf upgrade -y && \
     microdnf clean all
 
 ENV APP_ROOT=/opt/app-root


build/Dockerfile
```diff
@@ -4,6 +4,7 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1130 AS base
 
 RUN microdnf module enable nodejs:20 && \
     microdnf install -y shadow-utils jq nodejs --nodocs && \
+    microdnf upgrade -y && \
```
vkrizan, line 7: This would bloat the image. It its expected that the base would have these updates at the first place.

Are there any packages that we really need to update that are not updated on the ubi base?

vkrizan (commented):
I do not recommend adding this. See below.

build/Dockerfile
vkrizan: Ack. Feel free to merge. Thank you.

vkrizan (approved):

build/Dockerfile
rexwhite: The number of upgraded packages is minimal.  If this starts to cause bloat we can revisit it but as long as we're a current version of ubi-minimal this shouldn't be a problem.  Also, some of the updates are for packages installed as a result of installing node.

PR #571, opened 2024-11-21
RHINENG-14279: override execa to fix cross-spawn package vulnerability

rexwhite:
Added override for the execa pacakge to fix a cross-spawn vulnerability.


marleystipich2 (approved):
LGTM!

PR #575, opened 2024-12-04
RHINENG-14707: Reduce trace data for Patch issues

rexwhite:
Reduced the excessive amount of trace data we record for Patch issues.


marleystipich2 (approved):
LGTM!

PR #597, opened 2025-03-04
RHINENG-16337: sequence diagram for GET v1/remediations/:plan_id/playbook_runs/:run_id

rexwhite:
I added a sequence diagram for the v1/remediations/:plan_id/playbook_runs/:run_id endpoint.


marleystipich2 (approved):
Compared this with the code and it looks good to me too! 👍

PR #598, opened 2025-03-04
RHINENG-16365: Remove plan id as parameter to insights-client --diagnosis option

rexwhite:
The insights-client team removed support for specifying a "remediation plan id" with the --diagnosis option, which broke the remediations diagnosis special play, so I just removed the parameter from the call to insights-client (which ignores the parameter anyway and just gathers the pydata for ALL issues for the current system, so this change should not break anything).


marleystipich2 (approved):
That was fast! Looks good to me!

PR #613, opened 2025-03-27
Revert "RHINENG-16071: Adjust code to use Compliance API v2 endpoint"

rexwhite:
Reverts RedHatInsights/insights-remediations#599

A different approach has been suggested by the Compliance team.


marleystipich2:
(Thank you for doing this!! I was hoping we wouldn't need to but should've reverted before I left for PTO just in case)

PR #619, opened 2025-04-03
RHINENG-17235: Added missing 'additionalProperties: false' statement

rexwhite:
As a matter of policy we should explicitly list all properties allowed for an `object` and set `additionalProperties: false`


marleystipich2 (approved):
Makes sense! LGTM!

PR #627, opened 2025-04-16
RHINENG-16698: Added DELETE /remediations/:id/issues endpoint

rexwhite:
Added a bulk delete endpoint for issues: `DELETE /remediations/:id/issues`

### DIFF
--- a/src/remediations/__snapshots__/write.integration.js.snap
+++ b/src/remediations/__snapshots__/write.integration.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`remediations remove bulk delete empty list 1`] = `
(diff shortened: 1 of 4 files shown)


marleystipich2 (approved):
I tried to be like Jeremy and **thoroughly** review the test cases but I honestly think you covered it all!

One thing I thought of is when we're testing if a bulk delete of issues succeeded, we could verify that the remediation plan contains exactly the `issue_ids` that we expect after the delete. In the tests we're confirming that the `deleted_count` is correct but what if you added another line for checking the count of `issue_ids` at least or making sure the `issue_ids` match exactly what we expect. 

Totally just a suggestion! But approving this anyway 👍

rexwhite:
I'm including an update to that test in the next PR to avoid git rebase hell 😬

PR #668, opened 2025-05-19
fix(migrations): SQL syntax on fix-invalid-plan-names

vkrizan:
Fixes SQL in a migration to be compatible with PG 13.

## Summary by Sourcery

Bug Fixes:
- Add AS keyword for column alias in the SQL query within the migration


rexwhite (approved):
👍🏻

PR #678, opened 2025-06-11
RHINENG-18655: Added check for null return from db playbook run query

rexwhite:
Added a check for getPlaybookRuns returning a `null`.  I'm not entirely sure why this occasionally happens but suspect it may be a race condition between the dashboard widget and a user deleting a remediation plan 🤷🏻‍♂️

## Summary by Sourcery

Bug Fixes:
- Handle occasional null return from getPlaybookRuns to avoid runtime errors


marleystipich2 (approved):
Wooo! LGTM!

PR #691, opened 2025-07-09
Kessel poc

josejulio:
AI assisted PoC using the node sdk [@project-kessel/kessel-sdk](https://www.npmjs.com/package/@project-kessel/kessel-sdk) + documentation.

Did some changes on top of that but I haven't tested it.
This uses the default workspace pattern.

### DIFF
--- a/kessel-remediations.ksl
+++ b/kessel-remediations.ksl
@@ -1,0 +1,124 @@
+version 0.1
+namespace remediations
+
+import rbac
+import hbi
+
+// ========================================
+// Remediations v1 to v2 Permission Mappings
+// ========================================
+
+// Remediation Resource Permissions
+// --------------------------------
+
+// Maps remediations:remediation:read to workspace permission
+@rbac.add_v1_based_permission(app:'remediations', resource:'remediation', verb:'read', v2_perm:'remediations_read_remediation_assigned');
+
+// Maps remediations:remediation:write to workspace permission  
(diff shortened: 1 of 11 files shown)


kessel-remediations.ksl
```diff
@@ -0,0 +1,124 @@
...
+@hbi.expose_host_permission(v2_perm: 'remediations_write_playbook', host_perm: 'remediations_write_playbook');
+@hbi.expose_host_permission(v2_perm: 'remediations_execute_playbook', host_perm: 'remediations_execute_playbook');
+@hbi.expose_host_permission(v2_perm: 'remediations_read_system', host_perm: 'remediations_read_system');
+@hbi.expose_host_permission(v2_perm: 'remediations_write_system', host_perm: 'remediations_write_system');
+
+// ========================================
+// Wildcard Permission Mappings
+// ========================================
+
+// Handle wildcard permissions from v1 RBAC
```
wscalf, line 83: We shouldn't need these explicit wildcard mappings - add_v1_based_permission already creates the necessary wildcards and or's them in for the V2 permission. See rbac.ksl: https://github.com/RedHatInsights/rbac-config/blob/master/configs/stage/schemas/src/rbac.ksl#L53

josejulio:
Just a heads up, this is still an early PoC and requires more work. I can try to review/update what would need to be updated based on my experience with the other work I did.

PR #702, opened 2025-07-24
RHINENG-19556: Add dispatcher sync function

marleystipich2:
## Summary by Sourcery

Implement a dispatcher sync function in fifi_2.js that identifies which playbook runs need dispatcher data backfill or status updates, fetches data from the dispatcher API with concurrency limits, updates or inserts dispatcher_runs entries accordingly, and add comprehensive integration and unit tests.

New Features:
- Add syncDispatcherRunsForPlaybookRuns function to backfill missing dispatcher_runs and update statuses from dispatcher API for specified playbook runs

Enhancements:
- Limit concurrent dispatcher API calls using p-limit and log warnings on sync failures

Tests:
- Add integration tests for backfilling, status updates, and skipping failed runs in syncDispatcherRunsForPlaybookRuns
- Add unit tests covering empty input, backfill, status update, skip on failures, and API error handling

### DIFF
--- a/src/remediations/fifi_2.js
+++ b/src/remediations/fifi_2.js
@@ -10,6 +10,7 @@ const probes = require("../probes");
 const dispatcher = require("../connectors/dispatcher");
 const log = require("../util/log");
 const queries = require("./remediations.queries");
+const pLimit = require('p-limit');
 
 // status constants
 const CONNECTED = 'connected';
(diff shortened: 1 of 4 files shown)


marleystipich2:
One thing that this PR doesn't cover because it's not possible(I don't think): 
I looked into adding a `dispatcher-id` label to the response that we get from the playbook-dipatcher API call so that we can filter by dispatcher run ids using the label but this wasn't possible because at the time where we're sending work requests to playbook-dispatcher and adding labels, the dispatcher run id hasn't been created by them yet

Or maybe I'm missing something 🤔

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+            LEFT JOIN dispatcher_runs dr ON pr.id = dr.remediations_run_id
+            WHERE pr.id IN (:playbookRunIds)
+            GROUP BY pr.id
+        `;
+
+        const playbookRunCounts = await db.s.query(sql, {
+            replacements: { playbookRunIds },
+            type: db.s.QueryTypes.SELECT
+        });
+
```
rexwhite, line 253: Actually, the code from this point on kinda _does_ feel like it belongs in this file.  Maybe just move the db query part to `remediations.queries.js` and leave this part with the call to playbook-dispatcher here? 🤷🏻‍♂️

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+                try {
+                    const filter = {
+                        filter: { 
+                            service: 'remediations',
+                            labels: { 'playbook-run': playbookRunId } 
+                        }
+                    };
+                    const fields = {
+                        fields: { data: ['id', 'status'] }
+                    };
```
rexwhite, line 283: Ooh... the Sorcery comment above is actually pretty good.  Let me know if it's confusing (it's basically a way to return an array that only contains the runs that were successfully synced).

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+                                created_at: new Date(),
+                                updated_at: new Date(),
+                                pd_response_code: null
+                            }));
+                            
+                            await queries.insertDispatcherRuns(newRuns);
+                        } else {
+                            // Update existing dispatcher_runs entries
+                            for (const pdRun of dispatcherResponse.data) {
+                                await db.dispatcher_runs.update(
```
rexwhite, line 303: Don't do this - you're calling `queries.insertDispatcherRuns()` above to add new runs; add an `updateDispatcherRuns()` function to `queries` and call that instead of calling the model object directly.

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@
...
+                                            dispatcher_run_id: pdRun.id,
+                                            remediations_run_id: playbookRunId 
+                                        } 
+                                    }
+                                );
+                            }
+                        }
+                    }
+                } catch (error) {
+                    log.warn(`Failed to sync dispatcher data for playbook run ${playbookRunId}:`, error.message);
```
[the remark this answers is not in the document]
rexwhite: This is also not a bad idea...

rexwhite (approved):
This looks pretty good `functionally`, but I left some style / consistency suggestions.  Also, you might want to add a couple of comments just to describe what the function is doing at a high level in a few spots.

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+                // Has pending/running, needs update
+                needsStatusSync.push(runCounts.playbook_run_id);
+            }
+        }
+
+        const runsToSync = [...needsStatusSync, ...needsBackfill];
+
+        if (runsToSync.length > 0) {
+            // Limit concurrent API calls
+            const limit = pLimit(10);
```
rexwhite, line 271: I hadn't seen `p-limit` previously - it actually looks pretty useful!

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+            SELECT 
+                pr.id as playbook_run_id,
+                COUNT(dr.dispatcher_run_id)::int as total_dispatcher_runs,
+                COUNT(CASE WHEN dr.status IN ('failure', 'timeout') THEN 1 END)::int as failed_runs,
+                COUNT(CASE WHEN dr.status IN ('pending', 'running') THEN 1 END)::int as incomplete_runs
+            FROM playbook_runs pr
+            LEFT JOIN dispatcher_runs dr ON pr.id = dr.remediations_run_id
+            WHERE pr.id IN (:playbookRunIds)
+            GROUP BY pr.id
+        `;
```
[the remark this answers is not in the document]
marleystipich2: Ahh yeah I agree.. I asked Cursor to use SQL and told it what I wanted the query to do but I bet I could tell it to convert the raw SQL to Sequalize instead. I'll try that!

src/remediations/fifi_2.js
marleystipich2: Sourcery suggested this actually 😄

src/remediations/fifi_2.js
```diff
@@ -217,3 +219,112 @@ async function storePlaybookRun (playbook_run_id, remediation, username) {
...
+            await Promise.all(runsToSync.map(playbookRunId => limit(async () => {
+                try {
+                    const filter = {
+                        filter: { 
+                            service: 'remediations',
+                            labels: { 'playbook-run': playbookRunId } 
+                        }
+                    };
+                    const fields = {
+                        fields: { data: ['id', 'status'] }
```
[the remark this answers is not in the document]
marleystipich2: I made this change.

src/remediations/fifi_2.js
marleystipich2: Yep I like that too I made that change 🙂

PR #717, opened 2025-08-11
RHINENG-16541: Override form-data to 2.5.4

marleystipich2:
## Summary by Sourcery

Enhancements:
- Add an npm override for form-data to lock its version to 2.5.4 in package.json and update package-lock.json accordingly

### DIFF
--- a/src/connectors/compliance/impl.js
+++ b/src/connectors/compliance/impl.js
@@ -3,6 +3,7 @@
 const assert = require('assert');
 const _ = require('lodash');
 const log = require('../../util/log');
+const errors = require('../../errors');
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -74,7 +74,8 @@
     },
     "cls-hooked": {
       "semver": "^5.7.2"
-    }
+    },
+    "form-data": "2.5.4"
```
rexwhite, line 78: We might want to scope this to just `request` and see if we can bump `supertest` up to a newer versions (`supertest` is the other consumer of `form-data`).

rexwhite (approved):
Looks good, although we should eventually look at updating `supertest` to something less ancient.

PR #719, opened 2025-08-13
RHINENG-19924: Only allow v2 issue ids in SSGResolver

marleystipich2:


rexwhite (approved):
Looks good!  I like the descriptive message in the error response.

PR #741, opened 2025-09-10
Kessel PoC

brantleyr:
Initial Kessel PoC.
Based off of the original MR - https://github.com/RedHatInsights/insights-remediations/pull/691

### DIFF
--- a/docker-compose.kessel.yml
+++ b/docker-compose.kessel.yml
@@ -1,0 +1,50 @@
+version: '3.8'
+
+services:
+  # SpiceDB - the authorization database (managed by Kessel)
+  spicedb:
+    image: "authzed/spicedb:latest"
+    command: "serve"
+    restart: "always"
+    ports:
+      - "8080:8080"
+      - "9090:9090"
+    environment:
+      - SPICEDB_GRPC_PRESHARED_KEY=somerandomkeyhere
+      - SPICEDB_HTTP_ENABLED=true
+    depends_on:
+      - postgres
+
+  # PostgreSQL for SpiceDB
+  postgres:
+    image: "postgres:14"
+    ports:
+      - "5433:5432"
+    environment:
+      - POSTGRES_PASSWORD=secret
+      - POSTGRES_DB=spicedb
+      - POSTGRES_USER=spicedb
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+
+  # Kessel Relations API
+  kessel-relations:
+    image: "quay.io/cloudservices/kessel-relations:latest"
+    ports:
+      - "8081:8080"
+    environment:
+      - SPICEDB_ENDPOINT=spicedb:50051
+      - SPICEDB_TOKEN=somerandomkeyhere
+      - SPICEDB_INSECURE=true
+    depends_on:
+      - spicedb
+    restart: "always"
+
+  # Redis for caching (optional)
+  redis:
+    image: "redis:7-alpine"
+    ports:
+      - "6380:6379"
+
+volumes:
+  postgres_data: 
--- a/scripts/test-kessel-migration.js
+++ b/scripts/test-kessel-migration.js
@@ -1,0 +1,230 @@
+#!/usr/bin/env node
+
+'use strict';
+
+/**
+ * Test script for Kessel migration
+ *
+ * This script helps test both traditional RBAC and Kessel authorization
+ * to ensure the migration works correctly.
+ */
+
+const config = require('../src/config');
+const rbacConnector = require('../src/connectors/rbac');
+const kesselConnector = require('../src/connectors/kessel');
+
+async function testTraditionalRbac() {
+    console.log('\n=== Testing Traditional RBAC ===');
+
+    try {
+        console.log('RBAC Config:', {
+            impl: config.rbac.impl,
+            host: config.rbac.host,
+            enforce: config.rbac.enforce
+        });
+
+        if (config.rbac.enforce) {
+            console.log('Attempting to get RBAC access...');
+            const access = await rbacConnector.getRemediationsAccess();
+
+            if (access && access.data) {
+                console.log(`✓ RBAC Access retrieved: ${access.data.length} permissions found`);
+                access.data.forEach((perm, index) => {
+                    console.log(`  ${index + 1}. ${perm.permission}`);
+                });
+            } else {
+                console.log('⚠ RBAC Access returned null or empty');
+            }
+
+            console.log('Testing RBAC ping...');
+            await rbacConnector.ping();
+            console.log('✓ RBAC ping successful');
+        } else {
+            console.log('⚠ RBAC enforcement is disabled');
+        }
+    } catch (error) {
+        console.error('✗ Traditional RBAC test failed:', error.message);
+    }
+}
+
+async function testKessel() {
+    console.log('\n=== Testing Kessel ===');
+
+    try {
+        console.log('Kessel Config:', {
+            enabled: config.kessel.enabled,
+            url: config.kessel.url,
+            insecure: config.kessel.insecure,
+            timeout: config.kessel.timeout
+        });
+
+        if (config.kessel.enabled) {
+            console.log('Attempting to get Kessel access...');
+            const access = await kesselConnector.getRemediationsAccess();
+
+            if (access && access.data) {
+                console.log(`✓ Kessel Access retrieved: ${access.data.length} permissions found`);
+                access.data.forEach((perm, index) => {
+                    console.log(`  ${index + 1}. ${perm.permission}`);
+                });
+            } else {
+                console.log('⚠ Kessel Access returned null or empty');
+            }
+
+            console.log('Testing specific permission check...');
+            const hasPermission = await kesselConnector.hasPermission(
+                'remediation',
+                'read',
+                'test-user-123'
+            );
+            console.log(`✓ Permission check result (workspace-based): ${hasPermission}`);
+
+            console.log('Testing workspace permission mapping...');
+            const workspacePermission = kesselConnector.convertRbacToWorkspacePermission('remediation', 'read');
+            console.log(`✓ Workspace permission mapping: remediation:read → ${workspacePermission}`);
+
+            console.log('Testing Kessel ping...');
+            await kesselConnector.ping();
+            console.log('✓ Kessel ping successful');
+        } else {
+            console.log('⚠ Kessel is disabled');
+        }
+    } catch (error) {
+        console.error('✗ Kessel test failed:', error.message);
+    }
+}
+
+async function testRbacMiddleware() {
+    console.log('\n=== Testing RBAC Middleware Logic ===');
+
+    // Mock request object
+    const mockReq = {
+        identity: {
+            type: 'User',
+            identity: {
+                user: {
+                    user_id: 'test-user-123'
+                }
+            }
+        },
+        headers: {
+            'x-rh-identity': Buffer.from(JSON.stringify({
+                identity: {
+                    user: {
+                        user_id: 'test-user-123'
+                    }
+                }
+            })).toString('base64')
+        }
+    };
+
+    // Test different permission scenarios
+    const permissions = [
+        'remediations:remediation:read',
+        'remediations:remediation:write',
+        'remediations:playbook:execute',
+        'remediations:system:read'
+    ];
+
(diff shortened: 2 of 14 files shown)


src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,227 @@
...
+    async checkSinglePermission(userId, workspaceId, relation) {
+        try {
+            // Create the check request using the new API structure
+            const checkRequest = {
+                subject: {
+                    resource: {
+                        reporter: {
+                            type: "rbac"
+                        },
+                        resourceId: `redhat/${userId}`,
```
josejulio: Replace `redhat` with the value from the config: `KESSEL_PRINCIPAL_DOMAIN`

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,227 @@
...
+                return JSON.parse(identityHeader);
+            } catch (error) {
+                log.warn({ error }, 'Failed to parse x-rh-identity header');
+            }
+        }
+
+        return null;
+    }
+
+    async ping () {
```
josejulio: I don't think we really the `ping` and `pingPermissionCheck` functions - do you think we would need them?
I think it was initially suggested by claude, but probably not required.

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,227 @@
...
+
+            log.error({ error, resource, action, subject }, 'Failed to check permission with Kessel');
+            return false;
+        }
+    }
+
+    convertRbacToWorkspacePermission(resource, action) {
+        // Convert traditional RBAC permission to workspace permission
+        // Pattern: ${resource}:${action} -> remediations_${action}_${resource}
+        return `remediations_${action}_${resource}`;
```
josejulio: This will need some updates as well - we use different verbs e.g. `read` vs `view`

src/connectors/kessel/impl.js
brantleyr: These have now been removed.

scripts/test-kessel-migration.js
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env node
```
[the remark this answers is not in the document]
brantleyr: Where else do you think these tests should live?

docker-compose.kessel.yml
```diff
@@ -0,0 +1,50 @@
+version: '3.8'
```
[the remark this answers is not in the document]
josejulio: I think we can remove it and follow a similar approach to whatever you are currently doing to try with RBAC (some instructions perhaps?)

src/config/__snapshots__/config.unit.js.snap
```diff
@@ -160,7 +170,7 @@ exports[`Configuration Verify app-common-js function 1`] = `
   },
   "validateResponseStrict": true,
   "vmaas": {
-    "host": "https://console.redhat.com",
+    "host": "https://webapp-vmaas-prod.apps.crcp01ue1.o9m8.p1.openshiftapps.com",
```
josejulio: I don't recall changing this, just checking, if you did that's fine but if not, maybe we can revert it.

src/config/index.js
```diff
@@ -167,6 +167,17 @@ function Config() {
...
 
+        kessel: {
+            enabled: env.KESSEL_ENABLED === 'true' ? true : false,
+            authEnabled: env.KESSEL_AUTH_ENABLED === 'true' ? true : false,
+            oidcIssuerUrl: env.KESSEL_AUTH_OIDC_ISSUE,
+            clientId: env.KESSEL_AUTH_CLIENT_ID,
+            clientSecret: env.KESSEL_AUTH_CLIENT_SECRET,
+            url: env.KESSEL_URL || 'http://localhost:8080',
+            insecure: (env.KESSEL_INSECURE === 'true') ? true : false,
+            principalDomain: env.KESSEL_PRINCIPAL_DOMAIN || 'redhat'
```
josejulio: should we compare ignoring case for the booleans?

src/config/index.js
```diff
@@ -252,10 +263,11 @@ function Config() {
         config.inventory.host = getHostForApp(dependencyEndpoints, 'host-inventory', 'service', 'INVENTORY') || env.INVENTORY_HOST || 'http://insights-inventory.platform-ci.svc.cluster.local:8080';
         config.patchman.host = getHostForApp(dependencyEndpoints, 'patchman', 'manager', 'PATCHMAN') || env.PATCHMAN_HOST || 'http://localhost:8080';
         config.rbac.host = getHostForApp(dependencyEndpoints, 'rbac', 'service', 'RBAC') || env.RBAC_HOST || 'http://localhost:8080';
+        config.kessel.url = getHostForApp(dependencyEndpoints, 'kessel-relations', 'api', 'KESSEL_URL') || env.KESSEL_URL || 'http://localhost:8080';
```
josejulio: This should be setup already above and kessel is not being reported correctly by clowder for now

```suggestion
```

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,205 @@
...
+const metrics = require('../metrics');
+
+// Import new Kessel SDK with ClientBuilder
+let ClientBuilder, fetchOIDCDiscovery, OAuth2ClientCredentials, OAuth2AuthRequest, fetchDefaultWorkspace
+try {
+    // Import the new ClientBuilder from the updated SDK
+    const kesselSdk = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2');
+    ClientBuilder = kesselSdk.ClientBuilder;
+
+    const oAuthSdk = require('@project-kessel/kessel-sdk/kessel/auth');
```
josejulio: Lets import the Allowed enum, will be used below

```suggestion
let ClientBuilder, fetchOIDCDiscovery, OAuth2ClientCredentials, OAuth2AuthRequest, fetchDefaultWorkspace, Allowed
try {
    // Import the new ClientBuilder from the updated SDK
    const kesselSdk = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2');
    ClientBuilder = kesselSdk.ClientBuilder;
    
    const allowedSdk = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2/allowed');
    Allowed = allowedSdk.Allowed;

    const oAuthSdk = require('@project-kessel/kessel-sdk/kessel/auth');
```

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,205 @@
...
+                log.warn({ resource, action }, 'Unknown permission mapping');
+                return false;
+            }
+
+            // Get workspace ID from subject_org_id or use default
+            const workspace = await this.getDefaultWorkspaceIdForSubject(subject_org_id);
+            const workspaceId = workspace?.id || null;
+
+            // Check the specific permission using async/await
+            const allowed = await this.checkSinglePermission(subject_org_id, workspaceId, workspacePermission);
```
josejulio: Almost missed this.

We actually need the `user_id` - not the org_id here- You can extract it from the identity like so:

```
identity = the_json_x_rh_identity_header['identity'];
if (identity['type'] === 'ServiceAccount') {
  user_id = identity['service_account']['user_id']
} else if (identity['type'] === 'User') {
  user_id = identity['user']['user_id']
} else {
  throw new Error('unknown identity type:' + identity['type']);
}

org_id = identity['org_id']
```

brantleyr: I had used `org_id` here because it looked like the `fetchWorkspace` and `fetchDefaultWorkspace` functions from the SDK were using `orgId`.

- https://github.com/project-kessel/kessel-sdk-node/blob/main/src/kessel/rbac/v2.ts#L61

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,200 @@
...
+        try {
+            // Convert traditional RBAC permission to workspace permission
+            const workspacePermission = this.convertRbacToWorkspacePermission(resource, action);
+            if (!workspacePermission) {
+                log.warn({ resource, action }, 'Unknown permission mapping');
+                return false;
+            }
+
+            // Get workspace ID from subject_org_id or use default
+            const workspace = await this.getDefaultWorkspaceIdForSubject(subject_org_id);
```
josejulio: 💯 

This `org_id` usage is correct, you call `rbac` with the `org_id` to fetch the default `workspace id`

docker-compose.kessel.yml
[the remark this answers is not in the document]
brantleyr: I have removed this compose file and added instructions for the inventory-api repository.

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,205 @@
...
+                clientId: this.kesselConfig.clientId,
+                clientSecret: this.kesselConfig.clientSecret,
+                tokenEndpoint: discovery.tokenEndpoint,
+            });
+            const defaultWorkspace = await fetchDefaultWorkspace(
+                                    this.kesselConfig.url,
+                                    subject_org_id,
+                                    OAuth2AuthRequest(oAuth2ClientCredentials),
+                                );
+            return defaultWorkspace;
```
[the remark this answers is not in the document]
brantleyr: I have added this optimization.

src/connectors/kessel/impl.js
brantleyr: I have made those changes. Let me know if that resolves this issue.

src/middleware/rbac.js
```diff
@@ -6,39 +6,106 @@ const config = require('../config');
...
+        if (!identity) {
+            return false;
+        }
+
+        // Use Kessel connector to check permission
+        const hasPermission = await kesselConnector.hasPermission(
+            srcPermission.resource,
+            srcPermission.resourceType,
+            identity.org_id || identity.identity?.org_id,
+            identity.user_id || identity.identity?.user?.user_id
```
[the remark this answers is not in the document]
josejulio: I guess that might work

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,199 @@
+'use strict';
+
+const assert = require('assert');
+const Connector = require('../Connector');
+const log = require('../../util/log');
+
+const metrics = require('../metrics');
+
+// Import new Kessel SDK with ClientBuilder
+let ClientBuilder, fetchOIDCDiscovery, OAuth2ClientCredentials, OAuth2AuthRequest, fetchDefaultWorkspace, Allowed
```
rexwhite: ;

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,199 @@
...
+    const kesselSdk = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2');
+    ClientBuilder = kesselSdk.ClientBuilder;
+    
+    const allowedSdk = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2/allowed');
+    Allowed = allowedSdk.Allowed;
+
+    const oAuthSdk = require('@project-kessel/kessel-sdk/kessel/auth');
+    fetchOIDCDiscovery = oAuthSdk.fetchOIDCDiscovery;
+    OAuth2ClientCredentials = oAuthSdk.OAuth2ClientCredentials;
+    OAuth2AuthRequest = oAuthSdk.oauth2AuthRequest;
```
rexwhite: Super-annoying style comment you can totally disregard...

This is 100% fine but can be written more succinctly using destructuring:

```
const {ClientBuilder} = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2');
const {Allowed} = require('@project-kessel/kessel-sdk/kessel/inventory/v1beta2/allowed');
const {fetchOIDCDiscovery, OAuth2ClientCredentials, OAuth2AuthRequest} = require('@project-kessel/kessel-sdk/kessel/auth');
const {fetchDefaultWorkspace} = require('@project-kessel/kessel-sdk/kessel/rbac/v2');
```

src/connectors/kessel/impl.js
```diff
@@ -0,0 +1,199 @@
+'use strict';
```
rexwhite: We'll want to add prometheus probes to this at some point but we can do that later...

rexwhite (changes requested):
Mostly silly comments about missing semicolons and whatnot but DO take a look at the comment about the new connector returning a class rather than an object.

PR #1038, opened 2026-03-25
RHINENG-21568: Use our local systems table when generating playbooks

marleystipich2:
## Summary by Sourcery

Use the local systems table as the primary source for system details when generating playbooks, falling back to Inventory only for missing systems and caching those results locally.

New Features:
- Add a query to retrieve system details for playbook generation from the local systems table.
- Expose system detail storage for reuse when backfilling missing systems into the local database.

Enhancements:
- Update system resolution logic to combine local systems data with Inventory data and improve handling of missing systems.
- Extend seed data to include ansible_host for a test system to support the new system details flow.

### DIFF
--- a/src/generator/generator.controller.js
+++ b/src/generator/generator.controller.js
@@ -5,7 +5,9 @@ const P = require('bluebird');
 const etag = require('etag');
 
 const errors = require('../errors');
+const queries = require('../remediations/remediations.queries');
 const inventory = require('../connectors/inventory');
+const { storeSystemDetails } = require('../remediations/controller.write');
 const templates = require('../templates/static');
 const SpecialPlay = require('./plays/SpecialPlay');
 const format = require('./format');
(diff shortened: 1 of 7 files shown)


src/generator/generator.controller.js
```diff
@@ -146,40 +148,42 @@ exports.resolveSystems = async function (issues, strict = true) {
...
+        const inventoryData = await inventory.getSystemDetailsBatch(missingIds, true, 2, strict)
         .catch(e => {
             probes.failedGeneration('unknown system(s)');
             throw e;
         });
 
-    // When strict=false, filter out missing systems from issues
-    if (!strict) {
-        trace.event('Remove systems for which we have no inventory entry...');
-        _.forEach(issues, issue => issue.systems = issue.systems.filter((id) => {
```
marleystipich2, line 163: This code used to filter `issue.systems` to remove systems not found in Inventory but `issue.systems` is never used after this point so there's no reason to update it. Only `issue.hosts` is used so that's what needs to be filtered and contain the accurate systems and we're still doing that.

PR #1065, opened 2026-04-08
RHINENG-25171: Analyzed removal of npm from prod images

rexwhite:
## Summary by Sourcery

Documentation:
- Add a detailed design proposal outlining the removal of npm from production container images, including motivation, architecture, implementation plan, risk assessment, and rollout strategy.

### DIFF
--- a/plans/remove-npm/npm_removal_proposal.md
+++ b/plans/remove-npm/npm_removal_proposal.md
@@ -1,0 +1,394 @@
+# Proposal: Remove npm from Production Container Images
+
+**Date:** 2026-04-08  
+**Author:** Rex White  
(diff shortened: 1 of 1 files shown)


marleystipich2 (approved):
This sounds good to me and it's exactly what I had in mind too for this! 
One thing I mentioned on Slack but will add here too, we might need to update the `MIGRATION_COMMAND` line in our app-interface [deploy-clowder.yaml](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/insights/remediations/deploy-clowder.yml#L123) file? But also maybe not... Just something to keep in mind 🤷‍♀️

PR #1066, opened 2026-04-10
RHINENG-25338: Optimize formatRunHosts to get all hosts in 1 API call

marleystipich2:
## Summary by Sourcery

Optimize remediation run host formatting to retrieve all playbook run hosts in a single dispatcher API call and align mocks/tests with the new behavior.

Enhancements:
- Aggregate remediation hosts for a playbook run via a single dispatcher fetch filtered by playbook-run label and derive executor metadata from the associated run map.

Tests:
- Update and extend formatRunHosts unit tests to cover the new host payload shape, updated_at propagation, and playbook-run label handling in the dispatcher mock.

### DIFF
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -245,8 +245,8 @@ exports.formatRunHosts = async function (dispatcherRuns, playbook_run_id) {
             return hosts;
         }
 
-        // Collect all inventory IDs for batch system details lookup
-        const allInventoryIds = allRunHosts.data.map(host => host.inventory_id);
+        // Collect unique inventory IDs for batch system details lookup
+        const allInventoryIds = _.uniq(allRunHosts.data.map(host => host.inventory_id));
 
         // Fetch system details for all inventory IDs in batch
         const systemDetails = await queries.getPlanSystemsDetails(allInventoryIds);
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -26,7 +26,7 @@ const SYSTEM_FIELDS = Object.freeze(['id', 'ansible_host', 'hostname', 'display_
 
 const RUNSFIELDS = Object.freeze({fields: {data: ['id', 'labels', 'status', 'service', 'created_at', 'updated_at', 'url']}});
 const RUNHOSTFIELDS = Object.freeze({fields: {data: ['host', 'stdout', 'inventory_id']}});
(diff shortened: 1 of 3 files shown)


src/remediations/fifi.js
```diff
@@ -232,41 +232,39 @@ exports.formatRunHosts = async function (dispatcherRuns, playbook_run_id) {
...
+
+        if (!allRunHosts?.data) {
+            return hosts;
         }
 
+        // Collect all inventory IDs for batch system details lookup
+        const allInventoryIds = allRunHosts.data.map(host => host.inventory_id);
+
         // Fetch system details for all inventory IDs in batch
         const systemDetails = await queries.getPlanSystemsDetails(allInventoryIds);
```
rexwhite, line 252: Sourcery is right about this one - it would be better to remove duplicate systems before calling `getPlanSystemDetails()`.

src/remediations/fifi.js
```diff
@@ -232,41 +232,39 @@ exports.formatRunHosts = async function (dispatcherRuns, playbook_run_id) {
...
 
         // Format hosts with proper system names
-        hosts = allHosts.map(({ host, run, playbook_run_id }) => {
+        hosts = allRunHosts.data.map(host => {
             const details = systemDetails[host.inventory_id];
             // Use display_name if available, fallback to hostname, then to host.host
             const systemName = details?.display_name || details?.hostname || host.host;
             const isDirect = (host.host === 'localhost');
```
rexwhite, line 259: I know this was this way before but keying off of `host.host === 'localhost'` to determine if a host is directly connected is really unreliable - what if you have a satellite-managed host with a hostname of `localhost`?

We should probably (in a separate PR) add a column to the `dispatcher_runs` table to track the run type (compound satellite runs vs single-system direct runs) so we can reliably tell these two cases apart.

rexwhite (changes requested):
Since every dispatcher run will likely contain the same hosts, Sourcery's comment about deduplicating the host list before calling `getPlanSystemsDetails()` is correct.

src/remediations/fifi.js
marleystipich2: Yeah I agree and we actually are able to tell if the type is satellite from the Kafka messages that we receive because they only include `sat_id` and `sat_org_id`. I can open a ticket for this.

rexwhite (approved):
🤠👍🏻

PR #1073, opened 2026-04-16
RHINENG-25820: Fetch all run hosts in 1 call in formatRHCRuns

marleystipich2:
## Endpoints that use this function
- GET /remediations/{id}/playbook_runs
- GET /remediations/{id}/playbook_runs/{playbook_run_id}
- GET /remediations?fields=playbook_runs
- GET /remediations?fields=last_playbook_run

## Summary by Sourcery

Batch run host retrieval in RHC run formatting to reduce duplicate dispatcher calls.

Enhancements:
- Change RHC run formatting to fetch all hosts for a playbook run in a single dispatcher call and group them by run id before aggregating executor statistics.

Tests:
- Add unit coverage for combining dispatcher runs to ensure hosts are fetched once per playbook run and executor counts are correctly aggregated.

### DIFF
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -148,8 +148,8 @@ async function formatRHCRuns (dispatcherRuns, playbook_run_id) {
     for (const run of dispatcherRuns.data) {
         const hostList = hostsByRunId[run.id];
 
-        // For RHC direct there should be one run_host per dispatcher run (localhost) but still using _.some just to be safe.
-        if (_.some(hostList, ['host', 'localhost'])) {
+        // For RHC direct there should be one run_host per dispatcher run where host === 'localhost'
+        if (hostList?.length === 1 && hostList[0].host === 'localhost') {
             rhcDirect.playbook = run.url;
             rhcDirect.updated_at = run.updated_at;
             rhcDirect.system_count += hostList.length; // should always be 1, but...
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -138,25 +138,31 @@ async function formatRHCRuns (dispatcherRuns, playbook_run_id) {
 
     trace.event(`processing ${dispatcherRuns.data.length} runs...`);
 
+    // Fetch every run host for this playbook run from playbook-dispatcher
+    const runHostsFilter = createDispatcherRunHostsFilter(playbook_run_id);
+    const allRunHosts = await dispatcher.fetchPlaybookRunHosts(runHostsFilter, RHCRUNFIELDS);
+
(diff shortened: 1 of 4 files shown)


marleystipich2:
Note about this: We have plans to clean up `combineRuns` which calls `formatRHCRuns` so I think we should clean those up at the same time.. I was really trying to not do a full refactor here even though this function looks very confusing 😅 But this should definitely improve performance by a lotttt.

src/remediations/fifi.js
```diff
@@ -138,25 +138,31 @@ async function formatRHCRuns (dispatcherRuns, playbook_run_id) {
...
     for (const run of dispatcherRuns.data) {
-        // get dispatcher run hosts
-        const runHostsFilter = createDispatcherRunHostsFilter(run.labels['playbook-run'], run.id);
-        const rhcRunHosts = await dispatcher.fetchPlaybookRunHosts(runHostsFilter, RHCRUNFIELDS);
-        // If host === 'localhost' then add to RHCDirect
-        if (_.get(rhcRunHosts, 'data[0][host]') === 'localhost') {
+        const hostList = hostsByRunId[run.id];
+
+        // For RHC direct there should be one run_host per dispatcher run (localhost) but still using _.some just to be safe.
+        if (_.some(hostList, ['host', 'localhost'])) {
```
rexwhite: *Actually*... using `host === 'localhost'` to determine if this an RHC direct system is really unreliable and we should definitely use a better approach (in a different PR...).  It's actually *slightly* less error prone to say that any dispatcher run that has only one host, and that host's hostname is 'localhost', is likely an RHC direct run.  I get wanting to be safe here, but it's actually making it more likely to misidentify something.

Let's switch to checking if there's only one target in the run and it's `host === 'localhost'`

src/remediations/read.integration.js
```diff
@@ -530,6 +530,19 @@ describe('remediations', function () {
                     
                     // Use transaction for atomic setup
                     await db.s.transaction(async (transaction) => {
+                        // r249 (FiFI playbook 3): stray dispatcher_runs on any of its playbook runs change
```
rexwhite, line 533: Ugh... this is because our test seed data is a mess.  I really think we should only be using the seed data to test db migration edge cases (like, we add a new column and it's not populated for every row - we can use seed data to test that edge case handling).

This is fine for now, though...

rexwhite (changes requested):
There's one change request but otherwise this looks great - should have a huge impact!

src/remediations/read.integration.js
marleystipich2: Yep :/ The seed data is messy and then when other tests change things it affects other tests sometimes..

rexwhite (approved):
👍🏻

PR #1076, opened 2026-04-17
RHINENG-4887: Clean up and optimize getRunHostDetails (one run_hosts query)

marleystipich2:
## Extra summary explaining the refactoring/improvements 
How `getRunHostDetails` used to work:
1. Call `GET /v1/runs` using playbook-run label → all dispatcher runs for that remediation playbook run (could be dozens or hundreds)
2. Loop through each dispatcher run
3. For each dispatcher run, call `GET /v1/run_hosts` with playbook-run + run.id + inventory_id
4. If there's no data... that means there was no match for that playbook-run + run.id + inventory_id so keep looking... 
5. When you finally hit a match(`data` is not empty), you know that's the dispatcher run for that system
6. `formatRHCHostDetails(fullRun, rhcRunHosts, playbook_run_id)`

Performance before: 
- `GET /v1/run_hosts` - O(N)
In the worst case we had to call it once per dispatcher run until one returned a row for that system.
- `GET /v1/runs` - O(1)
We called it once, but that single response listed every dispatcher run for the playbook. 

New `getRunHostDetails` after this PR: 
1. Call `GET /v1/run_hosts` endpoint with playbook-run (remediation playbook_run_id) + inventory_id → at most one row; nested run.id is the dispatcher run id for that host
2. Read `dispatcherRunId` from `data[0].run.id`. If it’s missing, return `null` (no /runs call).
3. Check the local `dispatcher_runs` table for the dispatcher run (remediations_run_id = playbook_run_id, dispatcher_run_id = dispatcherRunId) 
4. If that row is missing (e.g. not synced yet), fallback: call `GET /v1/runs` with the playbook-run label, then `_.find` the dispatcher run whose id equals `dispatcherRunId`.
5. ` formatRHCHostDetails(fullRun, rhcRunHosts, playbook_run_id)`

Performance after: 
- `GET /v1/run_hosts` - O(1) - One request per lookup, filtered by playbook-run label and inventory_id. The response supplies the per-host payload (e.g. console output) and the nested run.id, which is the dispatcher run id for that host.
- `dispatcher_runs` (database) - Primary source for run-level fields needed for the API response. This avoids listing all dispatcher runs for the playbook when the row exists.
- `GET /v1/runs` (fallback) - O(1) - Used only when no dispatcher_runs row is found (e.g. not synced yet). 

## Summary by Sourcery

Optimize retrieval of run host details for a playbook run by querying run_hosts once per request and simplifying dispatcher run lookup.

Bug Fixes:
- Handle cases where run_hosts returns no data or entries missing run.id by returning null without querying dispatcher runs.

Enhancements:
- Include dispatcher run reference in run_host field selection to avoid multiple run_hosts queries.
- Refactor getRunHostDetails to derive the dispatcher run from run_hosts and then fetch the corresponding run record once.
- Improve tracing messages for clearer diagnostics around run_hosts and run lookups.

Tests:
- Add unit tests covering successful host detail retrieval and edge cases where run_hosts is empty or missing run.id.

### DIFF
--- a/src/remediations/fifi.js
+++ b/src/remediations/fifi.js
@@ -25,7 +25,7 @@ const MIN_SAT_RHC_VERSION = [6, 11, 0];
 const SYSTEM_FIELDS = Object.freeze(['id', 'ansible_host', 'hostname', 'display_name', 'rhc_client']);
 
 const RUNSFIELDS = Object.freeze({fields: {data: ['id', 'labels', 'status', 'service', 'created_at', 'updated_at', 'url']}});
-const RUNHOSTFIELDS = Object.freeze({fields: {data: ['host', 'stdout', 'inventory_id']}});
+const RUNHOSTFIELDS = Object.freeze({fields: {data: ['host', 'stdout', 'inventory_id', 'run']}});
 const RHCRUNFIELDS = Object.freeze({fields: {data: ['host', 'status', 'inventory_id', 'run']}});
 const RHCSTATUSES = ['timeout', 'failure', 'success', 'running', 'canceled'];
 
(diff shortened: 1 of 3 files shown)


src/remediations/fifi.js
```diff
@@ -341,52 +341,38 @@ exports.getRHCRuns = async function (playbook_run_id = null) {
...
+    const rhcRunHosts = await dispatcher.fetchPlaybookRunHosts(runHostsFilter, RUNHOSTFIELDS);
+    trace.event(`playbook-dispatcher/v1/run_hosts returned: ${JSON.stringify(rhcRunHosts)}`);
+
+    // rhcRunHosts will always have one row for the given playbook_run_id and system_id
+    // For direct: one dispatcher run per host 
+    // For satellite: many systems can share one dispatcher run but each system will still
+    // have its own run_host (same run.id, different inventory_id)
+    const dispatcherRunId = rhcRunHosts?.data?.[0]?.run?.id;
+    if (!dispatcherRunId) {
+        trace.leave(rhcRunHosts?.data?.length ? 'run_hosts row missing run.id' : 'data for system not found');
```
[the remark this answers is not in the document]
rexwhite: These items are actually included in the previous trace statement

src/remediations/fifi.js
```diff
@@ -341,52 +341,42 @@ exports.getRHCRuns = async function (playbook_run_id = null) {
...
-    const rhcRuns = await dispatcher.fetchPlaybookRuns(runsFilter, RUNSFIELDS);
-    trace.event(`playbook-dispatcher returned: ${JSON.stringify(rhcRuns)}`);
-
-    if (!rhcRuns || !rhcRuns.data) {
-        trace.leave('playbook-dispatcher returned nothing useful!');
-        return null; // didn't find any dispatcher runs for playbook_run_id...
+    // Fetch the run_host for the given playbook_run_id and system_id
+    const runHostsFilter = createDispatcherRunHostsFilter(playbook_run_id, null, system_id);
+    trace.event(`fetch playbook-dispatcher/v1/run_hosts with filter: ${JSON.stringify(runHostsFilter)}`);
+    const rhcRunHosts = await dispatcher.fetchPlaybookRunHosts(runHostsFilter, RUNHOSTFIELDS);
```
rexwhite, line 352: This is probably fine for now but at some point we should explore tracking the dispatcher_run_id per system in a plan_run_id so we don't have to call playbook-dispatcher to get this info.

src/remediations/fifi.js
```diff
@@ -341,52 +341,42 @@ exports.getRHCRuns = async function (playbook_run_id = null) {
...
-
-        if (rhcRunHosts.data) {
-            // there should only ever be one run_hosts entry for a given system_id in a
-            // dispatcher run, right?  Just grab the first entry...
-            const result = await exports.formatRHCHostDetails(run, rhcRunHosts, playbook_run_id);
-            trace.leave(`Found a match - returning: ${JSON.stringify(result)}`);
-            return result;
+    // Try to find the dispatcher run in the local dispatcher_runs table
+    let run = await queries.getDispatcherRunForPlaybookRun(playbook_run_id, dispatcherRunId);
+    if (!run) {
```
rexwhite, line 367: It feels like this code should be moved to some helper/wrapper function, but this is the only place it's used so 🤷🏻‍♂️

rexwhite (approved):
I left a couple comments but this is a great start for now!

PR #1086, opened 2026-04-22
RHINENG-24251: Removed config-manager checks

rexwhite:
## Summary by Sourcery

Remove usage of the config-manager connector and its RHC enablement checks from FiFi remediations and configuration.

Enhancements:
- Simplify FiFi connection status and playbook execution controllers by no longer checking RHC enablement via config-manager.
- Remove config-manager configuration sections and host resolution from the central config module.

Build:
- Drop config-manager related environment variables and parameters from deployment and clowdapp manifests.

Tests:
- Remove integration tests that assert 403 responses when RHC is disabled.
- Delete unit tests and snapshot content related to the config-manager connector.

Chores:
- Delete the config-manager connector implementation and associated mock files.


marleystipich2 (approved):
Looks good!! There are a couple skipped tests in `fifi.integration.js` that still stub configManager so we could remove those too but not super important since they're skipped 🤷‍♀️ 

But I verified that we won't be checking config manager anymore for these endpoints anymore (which is the most important part): 
- GET /v1/remediations/{id}/connection_status
- POST /v1/remediations/{id}/playbook_runs

PR #1095, opened 2026-04-28
RHINENG-26731 RHINENG-25793 RHINENG-25794: Add  /config endpoints

marleystipich2:
## Summary by Sourcery

Add organization-wide configuration API for plan retention settings with admin-protected updates and corresponding persistence.

New Features:
- Expose GET /config endpoint to retrieve organization plan retention configuration with sensible defaults.
- Expose PATCH /config endpoint to allow organization admins to set or update plan retention configuration.

Enhancements:
- Persist organization configuration in the org_config table via a dedicated controller and routes, including explicit table naming.
- Apply OpenAPI schemas for configuration input and response, including validation and defaulting semantics for retention and warning days.

Tests:
- Add unit tests for the configuration controller, including admin authorization and default resolution logic.
- Add integration tests for GET and PATCH /config endpoints covering authorization, validation, and persistence behavior.


src/api/openapi.yaml
```diff
@@ -765,6 +765,44 @@ paths:
         400:
           $ref: '#/components/responses/BadRequest'
 
+  /config:
```
rexwhite: Move this below the `resolutions` endpoints, but above the `version` endpoint, and create a new tag for it titled `configuration` with a one sentence description (see the other endpoint for tagging examples).  This will put it in its own section.

src/api/openapi.yaml
```diff
@@ -765,6 +765,44 @@ paths:
...
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ConfigResponse'
+    patch:
+      tags:
+        - remediations
+      summary: Update organization configuration
+      description: Updates organization-wide configuration for the current organization (organization admin only)
```
rexwhite: Change to: "Updates organization-wide configurable items for the caller's organization (restricted to organization admins only)"

src/api/openapi.yaml
```diff
@@ -1574,6 +1612,29 @@ components:
                 systems:
                   - a8799a02-8be9-11e8-9eb6-529269fb1459
 
+    ConfigInput:
+      type: object
+      additionalProperties: false
+      x-remediations-strict: false
+      description: >
+        Writable properties are plan_retention_days and plan_warning_days. When both are omitted,
```
rexwhite: What does this paragraph even mean?  I need to know:

1. How do I set one of these to a specific value,
2. how do I set it back to using the system default.
3. (ideally, what even *are* the systems defaults...)

"retained", "resolved values", "stored configuration", "service default" and "existing organizational value" wording is confusing.

src/api/openapi.yaml
```diff
@@ -1650,6 +1711,27 @@ components:
...
+      required: [data]
+      properties:
+        data:
+          type: object
+          additionalProperties: false
+          description: >
+            Values reflect stored organization configuration when available. If there are no stored values,
+            default values are plan_retention_days = 120, plan_warning_days = 30.
+          required: [plan_retention_days, plan_warning_days]
+          properties:
```
rexwhite, line 1847: 🤔 should we include the system defaults here?  Otherwise, how will a user know what they are?  Let's talk about this...

src/api/openapi.yaml
```diff
@@ -1574,6 +1612,29 @@ components:
...
+      x-remediations-strict: false
+      description: >
+        Writable properties are plan_retention_days and plan_warning_days. When both are omitted,
+        the resolved values (from stored configuration or service defaults) are persisted. If only one
+        property is sent, the other uses the existing organization value when set; otherwise service-wide
+        defaults apply.
+      properties:
+        plan_retention_days:
+          type: integer
+          minimum: 1
```
rexwhite, line 1738: Maybe `0 === system_default` for these?

src/api/openapi.yaml
```diff
@@ -1650,6 +1711,27 @@ components:
...
+      type: object
+      additionalProperties: false
+      description: Organization configuration for the current organization
+      required: [data]
+      properties:
+        data:
+          type: object
+          additionalProperties: false
+          description: >
+            Values reflect stored organization configuration when available. If there are no stored values,
```
rexwhite: This is misleading (and I hate the "stored organization configuration" wording).  This endpoint will _always_ return the two values, whether they are custom or default.  Also, how can the user tell if a custom value has been configured?  We need to talk about this a bit more

Also: we shouldn't list the default values in the api endpoint description like this because we'll change them and then forget they're mentioned here

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (user?.is_org_admin === true || serviceAccount?.is_org_admin === true) {
+        return next();
+    }
+
+    return next(new errors.Forbidden('Organization admin access required'));
+};
+
+exports.get = errors.async(async function (req, res) {
+    const org_id = req.user?.tenant_org_id || req.identity?.org_id;
```
rexwhite: Whaaaaat??  If `req.user.tenant_org_id` is nullish this would already have blown up in `src/middleware/identity/impl.js`.  The code is a little sloppy about validating that tenant_org_id is non-empty, but every endpoint needs that and if anything we should add that check to `src/middleware/identity/userIdentity.js`.  Perhaps I'm missing something but this line just makes no sense to me 🤷🏻‍♂️.

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+    const orgConfig = await db.org_config.findByPk(org_id);
+
+    return res.json(formatResponse(
```
rexwhite: I mentioned this elsewhere but I'm wondering if we shouldn't include the system default values here (or somewhere else)...

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
+'use strict';
+
+const errors = require('../errors');
+const db = require('../db');
+const config = require('../config');
+
+function formatResponse(plan_retention_days, plan_warning_days) {
```
rexwhite: Rename this `formatConfigResponse()` - there are so many `format...()` functions that in a stack trace it's hard to tell them apart.

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+    const orgConfig = await db.org_config.findByPk(org_id);
+
+    return res.json(formatResponse(
+        orgConfig?.plan_retention_days ?? config.plan_retention.retentionDays,
+        orgConfig?.plan_warning_days ?? config.plan_retention.warningDays
+    ));
+});
+
+exports.patch = errors.async(async function (req, res) {
+    const org_id = req.user?.tenant_org_id || req.identity?.org_id;
```
rexwhite: No.... see above

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+    const body = req.body ?? {};
+    const { plan_retention_days: bodyRetentionDays, plan_warning_days: bodyWarningDays } = body;
+
+    const existing = await db.org_config.findByPk(org_id);
+    const { retentionDays, warningDays } = config.plan_retention;
```
rexwhite: Make these: `defaultRetentionDays` and `defaultWarningDays`

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+
+    // For each field: use the request body when that property is not null or undefined,
+    // otherwise use the existing value in the org_config row,
+    // otherwise use config.plan_retention defaults.
+    const plan_retention_days = bodyRetentionDays ?? existing?.plan_retention_days ?? retentionDays;
+    const plan_warning_days = bodyWarningDays ?? existing?.plan_warning_days ?? warningDays;
+
+    await db.org_config.upsert({
```
rexwhite: I mentioned this elsewhere but with our current approach there's no way to "unset" these overrides once they're set...

src/config/config.controller.js
```diff
@@ -0,0 +1,57 @@
...
+function formatResponse(plan_retention_days, plan_warning_days) {
+    return {
+        data: {
+            plan_retention_days,
+            plan_warning_days
+        }
+    };
+}
+
+exports.requireOrgAdmin = function (req, res, next) {
```
rexwhite: This should probably be with the rest of the middleware.   Also, the global identity header processing code doesn't extract `is_org_admin` because this is the first endpoint to require it.  We should probably update `src/middleware/identity/impl.js` and handle this in a more proper fashion vs this one-off thing for just this endpoint

src/config/config.integration.js
```diff
@@ -0,0 +1,202 @@
+'use strict';
```
rexwhite, line 1: We'll probably want to addd "unset retention_days" tests once we decide how to handle that.

src/config/routes.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+const controller = require('./config.controller');
+const openapi = require('../middleware/openapi');
+
+module.exports = function (router) {
+    router.get('/config', openapi('getConfig'), controller.get);
+    router.patch('/config', controller.requireOrgAdmin, openapi('updateConfig'), controller.patch);
```
rexwhite: `requireOrgAdmin()` should probably live in '../middleware' somewhere

src/remediations/models/orgConfig.js
```diff
@@ -24,6 +24,7 @@ module.exports = (sequelize, {DATE, INTEGER, TEXT}) => {
             allowNull: false
         }
     }, {
+        tableName: 'org_config',
         timestamps: true,
         createdAt: 'created_at',
         updatedAt: 'updated_at'
```
rexwhite: Might want to add `deleted_at` if we decide to provide an `unset()` operation for these overrides.

rexwhite (changes requested):
Let's talk about these api additions - I think we should rethink this.

src/api/openapi.yaml
marleystipich2: Ahh yep I agree I didn't read through the descriptions close enough. I'll update these.

src/api/openapi.yaml
marleystipich2: Yep agreed! I updated the `ConfigResponse` to return default values. So it'll return:
```
- plan_retention_days
- plan_warning_days
- default_plan_retention_days
- default_plan_warning_days
```

src/config/config.controller.js
marleystipich2: We talked about this on a meeting but just replying to this anyway, I added the DELETE config/{field} endpoint and explained in the description that this will reset the field to it's default value so this will cover the case of "unsetting" the overrides

src/remediations/models/orgConfig.js
marleystipich2: So for the DELETE /config/{field} endpoint, I considered deleting the entry from the config table if all of the values are equal to the default values for the fields but I didn't do this because I thought it wouldn't hurt leaving the entry in the table 🤷‍♀️  But idk maybe there's a reason that we'd want to do this in the future?

src/api/openapi.yaml
marleystipich2: 😬 Yep re-wrote all of this

src/config/config.controller.js
marleystipich2: I updated this to `const { tenant_org_id } = req.user;` because we take care of the org id stuff in the middleware so this is enough and this is also what we do in the other controllers (for the most part).

src/config/config.controller.js
marleystipich2: I updated `src/middleware/identity/impl.js` and created an `orgAdmin` check in the middeware so we can re-use that in the future. I also adjusted the service account part in src/middleware/identity/impl.js like:
```
if (req.identity.type === 'ServiceAccount') {
            req.user = {
                account_number: '',
                tenant_org_id: req.identity.org_id,
                username: req.identity.service_account.username,
                is_internal: false,
                is_org_admin: req.identity.service_account.is_org_admin === true
            };
            req.type = "ServiceAccount";
}
```
Not sure if this is necessary 🤔

src/api/openapi.yaml
```diff
@@ -775,17 +782,37 @@ paths:
...
-                    default: 30
+                plan_retention_days:
+                  override: 90
+                  default: 120
+                plan_warning_days:
+                  override: null
+                  default: 30
         400:
-          $ref: '#/components/responses/BadRequest'
-        403:
```
marleystipich2: Did we remove the 403 response on purpose? Don't we return a 403 for non-admin users that aren't allowed to update org config values? Or maybe it's there and I'm not seeing it 🤔

rexwhite: Yeah, we don't typically explicitly list things like 403's or 500's - it's just understood that those are always possible.

rexwhite (approved):

PR #1096, opened 2026-04-29
RHINENG-24789: Adding proposal doc for using nodejs base image

marleystipich2:
## Summary by Sourcery

Documentation:
- Document a proposed approach for building and running the service on Hummingbird Node.js images, including version selection, multi-stage builds, hermetic Konflux usage, UID considerations, and operational tradeoffs.


rexwhite (approved):
You can merge this if you want, but it might be better to put this in a google doc since it's still in the "discuss / consider" stage.

PR #1101, opened 2026-05-04
RHINENG-26206: Add getSummary query func for format=summary endpoint

marleystipich2:
## Summary by Sourcery

Add a lightweight summary retrieval path for remediations and wire it to the existing read controller when format=summary is requested.

New Features:
- Introduce a getSummary query that returns remediation metadata and aggregated issue and system counts without loading full issue details.

Enhancements:
- Refine the remediation read controller to delegate summary responses to the new query instead of computing aggregates in-memory.

### DIFF
--- a/src/remediations/controller.read.js
+++ b/src/remediations/controller.read.js
@@ -316,6 +316,58 @@ function orderSystems (systems, column, asc = true) {
     return _.orderBy(systems, [column, 'id'], [asc ? 'asc' : 'desc']);
 }
 
+/**
+ * GET /v1/remediations/:id — sends `res.json(format.get(remediation))`.
+ *
+ * Default (`format` omitted or `detail`) loads full issues/systems and returns `needs_reboot`, `resolved_count`,
+ * and `issues`. `?format=summary` uses `queries.getSummary` and returns counts only (no `issues`).
+ *
+ * @example Summary (`?format=summary`) response body
+ * {
+ *   "id": "9197ba55-0abc-4028-9bbe-269e530f8bd5",
+ *   "name": "Fix Critical CVEs",
+ *   "auto_reboot": true,
+ *   "archived": false,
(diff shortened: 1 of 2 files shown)


src/remediations/remediations.queries.js
```diff
@@ -466,6 +466,68 @@ exports.get = async function (id, tenant_org_id, created_by = null, includeResol
...
 
+/**
+ * Gets counts and remediation metadata for `GET /v1/remediations/:id?format=summary`.
+ *
+ * Loads the remediation row plus aggregates: total issue count, distinct system count across those issues,
+ * and per-issue-type counts (issue_id prefix before `:`).
+ *
+ * Lightweight because it does not load issues or issue-system rows.
+ */
+exports.getSummary = async function (id, tenant_org_id, created_by = null) {
```
rexwhite, line 499: Add an example of the object returned so I don't have to run this in the debugger to figure out what it looks like.

src/remediations/controller.read.js
rexwhite, line 371: I know it didn't have one before but we really should add an example of what this function returns too, otherwise it's hard to tell without asking AI or running it in the debugger...

src/remediations/controller.read.js
```diff
@@ -325,7 +325,9 @@ exports.get = errors.async(async function (req, res) {
     // otherwise use creator
     const creator_sa_filter = req.type == "ServiceAccount" ? null : req.user.username;
 
-    let remediation = await queries.get(req.params.id, req.user.tenant_org_id, creator_sa_filter);
+    const remediation = summarize
```
rexwhite, line 380: I'm not wild about this: `remediation` contains different things depending on the value of `summarize`, which makes the code harder to follow.

That said, the whole `?format=summary` thing is already a bit of a hack; we really just need to redesign the whole api and publish a v2 version...

But that's for another sprint. 😬

rexwhite (approved):
Approved, but please add sample output to the two functions I mentioned so things are a little easier to follow 🙏🏻

PR #1129, opened 2026-05-18
RHINENG-26507: Replace uuid dependency with crypto.randomUUID

marleystipich2:
## Summary by Sourcery

Replace usage of the external uuid library with Node's built-in crypto.randomUUID across remediations code, middleware, tests, and helpers, and adjust related test utilities accordingly.

Enhancements:
- Standardize UUID generation via a new generatePlaybookRunId helper that wraps crypto.randomUUID for playbook runs.
- Simplify request ID middleware and test helpers to rely on crypto.randomUUID instead of an external dependency.

Build:
- Remove the uuid dependency from package.json now that UUID generation uses the built-in crypto module.

### DIFF
--- a/db/seeders/20240326162729-scalability.js
+++ b/db/seeders/20240326162729-scalability.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const _ = require('lodash');
-const UUID = require('uuid');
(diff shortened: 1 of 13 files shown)


rexwhite (approved):
`uuid.v4` was a little more informative (if you cared that we're using v4 uuids, which... meh 🤷🏻‍♂️), otherwise this looks fine.

You have a merge conflict with `package.json` though...

PR #1158, opened 2026-06-03
RHINENG-26731: Update config endpoints

marleystipich2:
## Summary by Sourcery

Revise configuration APIs to separate effective values, system defaults, and organization-specific overrides, simplifying response shapes and replacing patch/delete operations with a single override-focused endpoint.

New Features:
- Introduce endpoints to retrieve system default configuration values and organization-specific override values separately from effective values.

Enhancements:
- Change GET /config to return effective configuration values directly instead of nested override/default objects.
- Replace PATCH /config and DELETE /config/{field} with PUT /config/overrides, which clears omitted fields to defaults and returns only non-default overrides.
- Update OpenAPI specifications, schemas, and examples to reflect the new configuration model and validation behavior.

Tests:
- Update and expand unit and integration tests to cover the new configuration endpoints, override semantics, validation scenarios, and response formats.

Made the updates needed for this new design:

method & resource | request | response
-- | -- | --
GET /config | n/a | {"a": 30, "b": 120}
GET /config/defaults | n/a | {"a": 30, "b": 60}
GET /config/overrides | n/a | {"b": 120}
PUT /config/overrides | {"b": 90} | {"b": 90}

## Summary by Sourcery

Revise configuration API to return effective values directly and introduce separate endpoints for system defaults and organization-specific overrides, replacing patch/delete operations with a single override-focused update endpoint.

New Features:
- Add GET /config/defaults endpoint to return system default configuration values.
- Add GET /config/overrides endpoint to return only organization-specific override values.
- Add PUT /config/overrides endpoint to replace configuration overrides for an organization.

Enhancements:
- Change GET /config to return effective configuration values without nested override/default objects.
- Adjust override semantics so omitted fields are cleared back to defaults and responses only include non-default overrides.
- Update OpenAPI schemas, descriptions, and examples to reflect the new configuration model and validation behavior.

Tests:
- Update unit and integration tests to cover the new endpoints, override-clearing semantics, and validation rules while removing delete/patch-specific tests.

### DIFF
--- a/src/api/openapi.yaml
+++ b/src/api/openapi.yaml
@@ -727,13 +727,15 @@ paths:
         - configuration
       summary: Get organization-wide configuration
       description: |
-        Returns the current and default configuration values for the caller's organization.
+        Returns the effective value of each organization-wide configuration field for the
+        caller's organization.
 
-        Each field has an `override` value (custom setting) and a `default` value (system default).
-        - `override: null` means no custom value is set; the system default is used
-        - `override: <value>` means a custom value has been explicitly configured
+        Each field reflects a custom override value when one has been set, or the system default
+        value when no override exists. All configuration fields are always included in the response.
 
-        The effective value for a field is: `override ?? default`
+        To see only the system default value for each field, use **GET /config/defaults**.
+
+        To see only fields with custom override values, use **GET /config/overrides**.
       operationId: getConfig
       responses:
         200:
@@ -743,37 +745,16 @@ paths:
               schema:
                 $ref: '#/components/schemas/ConfigResponse'
               example:
-                plan_retention_days:
-                  override: null
-                  default: 120
-                plan_warning_days:
-                  override: 14
-                  default: 30
-    patch:
+                plan_retention_days: 120
+                plan_warning_days: 14
+
+  /config/defaults:
+    get:
       tags:
         - configuration
-      summary: Update organization configuration
-      description: |
-        Updates organization-wide configurable items for the caller's organization (restricted to organization admins only).
-
-        Sets custom values that override the system defaults. Use DELETE /config/{field} to reset a field back to its default value.  Alternately, send `null` to clear an override.
-
-        **Valid ranges:**
-        - `plan_retention_days`: 1-365 days (or `null` to use default)
-        - `plan_warning_days`: 1-365 days (or `null` to use default)
-
-        **Validation rules:**
-        - `plan_warning_days` must be less than `plan_retention_days` (using effective values after resolving nulls to system defaults)
-      operationId: updateConfig
-      requestBody:
-        required: true
-        content:
-          application/json:
-            schema:
-              $ref: '#/components/schemas/ConfigInput'
-            example:
-              plan_retention_days: 90
-              plan_warning_days: null
+      summary: Get system default values for organization-wide configuration fields
+      description: Returns the system default value for each organization-wide configuration field.
+      operationId: getConfigDefaults
       responses:
         200:
           description: OK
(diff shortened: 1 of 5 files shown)


marleystipich2:
@sourcery-ai review

src/api/openapi.yaml
```diff
@@ -743,37 +745,88 @@ paths:
...
+      summary: Replace organization-wide configuration overrides
+      description: |
+        Replaces override values for organization-wide configuration fields (restricted to
+        organization admins only).
+
+        Fields omitted from the request body, or set to `null`, revert to the system default (stored as null).
+        To keep an existing override unchanged when updating another field, include that field in the request 
+        body with its current override value. To reset all overrides, send an empty object `{}`.
+
+        Only fields with a custom override value are included in the response.
```
marleystipich2, line 804: For the `PUT /config/overrides` response, do we only want to return the fields with overrides/were included in the request body? That’s how I currently have it setup. 

So if request body is `{"a": 120, "b": 90}` then response is `{"a": 120, "b": 90}`

But if you want to reset all fields to their default values by using the request body `{}` then the response would also be `{}` . 

This endpoint is technically for updating override values so it make sense to only return the fields that have overrides.

rexwhite: I think both ways would return the same thing, since if you don't include a field in the PUT body it is removed - on a successful PUT the new set of overrides will exactly match the PUT body.

src/api/openapi.yaml
```diff
@@ -743,37 +745,88 @@ paths:
...
       requestBody:
         required: true
         content:
           application/json:
             schema:
               $ref: '#/components/schemas/ConfigInput'
-            example:
-              plan_retention_days: 90
-              plan_warning_days: null
+            examples:
```
rexwhite, line 820: I didn't realize you could have multiple examples 🙄

src/config/config.controller.js
```diff
@@ -5,144 +5,96 @@ const db = require('../db');
...
- *     "override": null,     // null = using system default
- *     "default": 120
- *   },
- *   "plan_warning_days": {
- *     "override": 14,       // custom value set
- *     "default": 30
- *   }
- * }
- */
 exports.get = errors.async(async function (req, res) {
```
rexwhite, line 41: Add a comment to describe this function, with sample output.

src/api/openapi.yaml
```diff
@@ -1725,30 +1722,31 @@ components:
...
-        - Set to a value (1-365) to create/update a custom override
+        **Important:** Fields omitted from the request body, or set to `null`, revert to the system default (stored as null).
+        To keep an existing override unchanged when updating another field, include that field in the request 
+        body with its current override value. To reset all overrides, send an empty object `{}`.
       properties:
         plan_retention_days:
           type: integer
           nullable: true
           minimum: 1
           maximum: 365
```
rexwhite: This should *probably* be 366, to allow for leap years 😬

src/api/openapi.yaml
```diff
@@ -1725,30 +1722,31 @@ components:
...
-            Set to null to use system default (120 days). Set to 1-365 to override.
+            Organization-wide configuration field. If a plan isn't updated or executed within this
+            many days, it will be treated as expired. Omit this field from the request body or set
+            to `null` to use the system default. Must be greater than `plan_warning_days`.
           example: 90
         plan_warning_days:
           type: integer
           nullable: true
           minimum: 1
           maximum: 365
```
rexwhite: 366? (See above)

rexwhite (approved):
Looks good.  Maybe add a comment for GET /config...

src/api/openapi.yaml
marleystipich2: Same lol Cursor figured it out for me 😄

PR #1162, opened 2026-06-09
RHINENG-26902: add subscriptionRefresh play for patch remediations

rverdile:
This supports a feature of Patch that will pre-calculate template advisory installability. This means that Patch may know that an advisory is installable before a system does. In the case where a user switches a system from one template to another, this could break a patch remediation if the advisory is installable according to patch, but the system has not been refreshed recently and therefore is not actually installable on the system.

I added a special play that does a subscription-manager refresh before dnf update. It only adds the refresh on template-managed systems. I also added a mock template managed system, so it can be verified locally with:

```
curl -s -X POST "$BASE_URL" \
  -H 'Content-Type: application/json' \
  -H "x-rh-identity: $IDENTITY" \
  -d '{"issues":[{"id":"patch-advisory:RHBA-2019:4105","systems":["32801e58-5765-4a0e-b1a0-1ab226aafeaa"]}]}'
```

Happy to answer any questions. Also happy to change anything or everything. I did my best to follow existing precedence, but this seemed like a weird case, and I've never worked on this project before :)

## Summary by Sourcery

Add a subscription-manager refresh play to patch remediation playbooks for template-managed systems and wire it through the Patchman connector.

New Features:
- Generate a dedicated subscription refresh play in remediation playbooks for template-managed systems before package updates.
- Expose a Patchman connector API to fetch template-managed system IDs and use it when building playbooks.
- Provide a static Ansible template for the subscription-manager refresh play.

Enhancements:
- Integrate template-managed system detection into the playbook generation pipeline while gracefully degrading on Patchman API failures.

Tests:
- Add integration tests verifying subscription refresh plays are added only for template-managed systems and skipped on failures or non-template systems.
- Add unit tests for the Patchman connector method that retrieves template-managed system IDs.

### DIFF
--- a/src/connectors/patchman/impl.js
+++ b/src/connectors/patchman/impl.js
@@ -12,6 +12,7 @@ module.exports = new class extends Connector {
     constructor () {
         super(module);
         this.metrics = metrics.createConnectorMetric(this.getName());
+        this.templateMetrics = metrics.createConnectorMetric(this.getName(), 'getTemplateSystemIds');
     }
 
     async getErratum (id, refresh = false) {
@@ -40,6 +41,26 @@ module.exports = new class extends Connector {
 
             throw e;
         }
+    }
+
+    async getTemplateSystemIds (systemIds) {
+        const uri = new URI(host);
+        uri.path('/api/patch/v3/systems');
+        uri.search({
+            'filter[id]': `in:${systemIds.join(',')}`,
+            'filter[template_uuid]': 'neq:null',
+            limit: -1
+        });
+
+        const res = await this.doHttp({
+            uri: uri.toString(),
+            method: 'GET',
+            json: true,
+            headers: this.getForwardedHeaders()
+        }, false, this.templateMetrics);
+
+        const ids = _.get(res, ['data'], []).map(system => system.id);
+        return new Set(ids);
     }
 
     async ping () {
--- a/src/connectors/patchman/mock.js
+++ b/src/connectors/patchman/mock.js
@@ -5,6 +5,10 @@ const Connector = require('../Connector');
 
 /* eslint-disable security/detect-object-injection, max-len */
 
+const TEMPLATE_MANAGED_SYSTEMS = new Set([
+    '32801e58-5765-4a0e-b1a0-1ab226aafeaa'
+]);
+
 const ERRATA = {
     'RHBA-2019:4105': {
         attributes: {
@@ -69,6 +73,11 @@ module.exports = new class extends Connector {
 
     getErratum (id) {
         return P.resolve(ERRATA[id]);
+    }
+
+    getTemplateSystemIds (systemIds) {
+        const result = new Set(systemIds.filter(id => TEMPLATE_MANAGED_SYSTEMS.has(id)));
+        return P.resolve(result);
     }
 
     ping () {
--- a/src/generator/generator.controller.js
+++ b/src/generator/generator.controller.js
@@ -11,6 +11,7 @@ const SpecialPlay = require('./plays/SpecialPlay');
 const format = require('./format');
 const identifiers = require('../util/identifiers');
 const erratumPlayAggregator = require('./erratumPlayAggregator');
+const patchman = require('../connectors/patchman');
 const issueManager = require('../issues');
 const log = require('../util/log');
 const trace = require('../util/trace');
@@ -38,6 +39,26 @@ exports.playbookPipeline = async function ({issues, auto_reboot = true}, remedia
     _.forEach(issues, issue => {
         issue.id = identifiers.parse(issue.id);
         trace.event(`issue.id = ${issue.id}`);
+    });
+
+    const patchmanSystemIds = _(issues)
+        .filter(i => i.id.app === 'patch-advisory')
+        .flatMap('systems')
+        .uniq()
+        .value();
+
+    const templateSystemIdsPromise = patchmanSystemIds.length > 0
+        ? patchman.getTemplateSystemIds(patchmanSystemIds).catch(e => {
+            log.warn(e, 'failed to query patchman for template systems, skipping subscription refresh play');
+            return new Set();
+        })
+        : Promise.resolve(new Set());
+
+    const systemIdToHostMap = new Map();
+    issues.forEach(issue => {
+        issue.systems.forEach((systemId, index) => {
+            systemIdToHostMap.set(systemId, issue.hosts[index]);
+        });
     });
 
     trace.event('Get play snippets for each issue...');
@@ -77,7 +98,12 @@ exports.playbookPipeline = async function ({issues, auto_reboot = true}, remedia
     trace.event('Aggregate erratum plays...');
     issues = erratumPlayAggregator.process(issues);
 
-    // Add play that generates a new Compliance report when there are Compliance(ssg) issues  
+    if (patchmanSystemIds.length > 0) {
+        trace.event('Add subscription-manager refresh play...');
+        issues = await addSubscriptionRefreshPlay(issues, templateSystemIdsPromise, systemIdToHostMap);
+    }
+
+    // Add play that generates a new Compliance report when there are Compliance(ssg) issues
     const complianceIssue = _.some(issues, issue => issue.id.app === 'ssg');
     if (complianceIssue) {
         trace.event('Generate new Compliance report...');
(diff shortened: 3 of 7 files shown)


src/connectors/patchman/impl.js
```diff
@@ -12,6 +12,7 @@ module.exports = new class extends Connector {
     constructor () {
         super(module);
         this.metrics = metrics.createConnectorMetric(this.getName());
+        this.templateMetrics = metrics.createConnectorMetric(this.getName(), 'getTemplateSystemIds');
```
rverdile, line 15: I'm wasn't sure how these metrics are used so I added a new one for this API call. I could remove this and use the other one if that doesn't make sense.

src/connectors/patchman/mock.js
```diff
@@ -5,6 +5,10 @@ const Connector = require('../Connector');
 
 /* eslint-disable security/detect-object-injection, max-len */
 
+const TEMPLATE_MANAGED_SYSTEMS = new Set([
+    '32801e58-5765-4a0e-b1a0-1ab226aafeaa'
+]);
+
```
rverdile, line 11: It doesn't seem to follow precedent, but I added this mock here so I could test the curl command to generate the playbook locally

src/templates/static/special/subscriptionRefresh.yml
```diff
@@ -0,0 +1,11 @@
+- name: refresh subscription-manager
+  hosts: "@@HOSTS@@"
+  become: true
+  gather_facts: false
+  vars:
+    insights_signature_exclude: /hosts,/vars/insights_signature
+    insights_signature: ""
```
rverdile, line 7: I see this needs a signature added, but I don't know what the process is there

src/generator/generator.controller.js
```diff
@@ -40,6 +41,18 @@ exports.playbookPipeline = async function ({issues, auto_reboot = true}, remedia
...
+        .filter(i => i.id.app === 'patch-advisory')
+        .flatMap('systems')
+        .uniq()
+        .value();
+    const systemIdToHostMap = new Map();
+    issues.forEach(issue => {
+        issue.systems.forEach((systemId, index) => {
+            systemIdToHostMap.set(systemId, issue.hosts[index]);
+        });
+    });
```
rverdile, line 62: I did this sort of mapping the id in the controller so that the hostnames could be passed to the play. It may have also made sense to do the patch lookups in the play itself, but it changing it this way limited the impact to the other plays. Not sure what the best approach was here.

src/connectors/patchman/impl.js
rexwhite: These metrics are sent to Prometheus, and then Grafana uses the info to display in dashboard widgets.  Adding these probes/metrics doesn't hurt anything but without updating Grafana they don't do any good either.

We are going to rework our Grafana dashboard in a separate effort and will review all of these metrics then.  For now this is totally fine, though 🙂.

src/templates/static/special/subscriptionRefresh.yml
rexwhite: I'll sign it for you once I review everything.

marleystipich2:
I reviewed these changes and overall I think they look great! There are certain things that might introduce performance problems so we'll probably want to make some adjustments before merging. I'm going to leave a couple comments for now just so I don't forget and think about this more next week. Good job figuring out this confusing codebase!! 😅  @rverdile

src/generator/generator.controller.js
```diff
@@ -228,6 +246,37 @@ function addDiagnosisPlay (plays, remediation = false) {
...
 }
 
+async function addSubscriptionRefreshPlay (plays, patchmanSystemIds, systemIdToHostMap) {
+    if (patchmanSystemIds.length === 0) {
+        return plays;
+    }
+
+    let templateSystemIds;
+    try {
+        templateSystemIds = await patchman.getTemplateSystemIds(patchmanSystemIds);
```
marleystipich2: I think we can improve performance if we adjust how/when we're using the Promise/await stuff for this. We get the `patchmanSystemIds` around line 48 in the `playbookPipeline` function so we have the info that we need to start this call at that time. 

The way that this is set up currently, we get `patchmanSystemIds` then we make the `createPlay` call and wait for that to finish and then later in the function we make the `addSubscriptionRefreshPlay` call using `patchmanSystemIds` that we got earlier in the function. It could help with performance if we started the `getTemplateSystemIds` call as soon as we have `patchmanSystemIds` but we can use Promise/await so that we aren't blocking the `createPlay` call that happens later. So basically these 2 calls could run in parallel instead of one after the other.

rverdile: thanks @marleystipich2 ! I pushed that change

src/connectors/patchman/impl.js
```diff
@@ -42,6 +43,26 @@ module.exports = new class extends Connector {
         }
     }
 
+    async getTemplateSystemIds (systemIds) {
+        const uri = new URI(host);
+        uri.path('/api/patch/v3/systems');
+        uri.search({
+            'filter[id]': `in:${systemIds.join(',')}`,
```
rexwhite, line 50: This should be chunked, otherwise the constructed url can exceed various limits.

src/connectors/patchman/mock.js
rexwhite: Yeah, this is fine.  I started migrating to a system where parts of a system UUID control various aspect of the system -- like belonging to a satellite, being disconnected, etc -- thinking that would make it easier to create specific systems for various tests but it turns out to be just as much work.  If we ever convert all the tests to use this approach we can update this mock at that time 🙂.

src/generator/generator.controller.js
```diff
@@ -40,6 +41,26 @@ exports.playbookPipeline = async function ({issues, auto_reboot = true}, remedia
         trace.event(`issue.id = ${issue.id}`);
     });
 
+    const patchmanSystemIds = _(issues)
```
rexwhite, line 44: Yeah, I kinda hate this... but it works.  These sorts of shenanigans are necessary because at the start of playbookPipeline() `issues` is an array of issues and affected systems, but by the end of the function `issues` is an array of plays 🤮.

I really don't want to pile more gross code on top of the pile of existing gross code.  I see three options:

1. We can accept this PR as-is and then refactor `playbookPipeline()` to be something more sane - I'm not a fan of this option strictly because while we *intend* to refactor it, it'll likely be some time before we get around to it.
2. You can refactor it yourself.  I hate asking you to clean up our gross code, though.  Plus this is a really performance-sensitive part of the code and I don't want to be critical of your re-write 😬...
3. I can do the refactor myself, lift the new parts of code you've written, submit a new PR, and then let you and Marley review it.  This would also be the easiest way for me to sign the new special play.

 (Also, it would be cleaner if we moved this code into `addSubscriptionRefreshPlay()`)

Do you have any thoughts?

src/generator/generator.controller.js
```diff
@@ -228,6 +254,27 @@ function addDiagnosisPlay (plays, remediation = false) {
...
+async function addSubscriptionRefreshPlay (plays, templateSystemIdsPromise, systemIdToHostMap) {
+    const templateSystemIds = await templateSystemIdsPromise;
+
+    if (templateSystemIds.size === 0) {
+        return plays;
+    }
+
+    const hosts = _(Array.from(templateSystemIds))
+        .map(id => systemIdToHostMap.get(id))
+        .filter()
```
rexwhite, line 266: So, lodash's compact() is faster if you're just trying to filter out the nullish items, and this part of the code is time-critical...

src/generator/generator.controller.js
```diff
@@ -228,6 +254,27 @@ function addDiagnosisPlay (plays, remediation = false) {
...
+        .uniq()
+        .sort()
```
rexwhite, line 268: Do we really need the list sorted?  This section of the code needs to be as efficient as possible...

rexwhite (changes requested):

src/generator/generator.controller.js
rverdile: My only goal is to get the spirit of this change in, so that we can finish the feature we're working on for Patch. I'm happy either try and refactor or let you refactor. I don't want to implement this in a gross / wrong way :)

I guess the one question I have is, if we choose to refactor this now, is that something you guys would be able to focus on?

And what do you prefer? I can try to refactor it; I'm not afraid of feedback. I don't know how big the refactor would be. I suspect that would require a lot of input from you. I don't know if that's actually any easier than you taking over and refactoring it. I'm willing to do this whichever way is easier for you guys.

rverdile: In the meantime, I'll cleanup the other smaller issues you've pointed out!

PR #1172, opened 2026-06-17
RHINENG-25797: Add upper limit for expires_within filter

marleystipich2:
## Summary by Sourcery

Enforce and document an upper bound on the expires_within remediation filter and update validation tests accordingly.

Bug Fixes:
- Validate filter.expires_within server-side to ensure it is an integer within the allowed range.

Enhancements:
- Clarify OpenAPI documentation for the expires_within filter to specify the valid numeric range and behavior.

Tests:
- Adjust and expand integration tests for the expires_within filter to cover the new maximum and invalid values.


rexwhite (approved):
It seems like Remediations has plenty of code to work around deficiencies with our OpenAPI validator - perhaps we could switch to a better option at some point.  For now, though, this looks good.  🙂

PR #1173, opened 2026-06-17
RHINENG-21568: Handle HBI 404s and use local systems table for playbooks

marleystipich2:
## Summary by Sourcery

Resolve systems for playbook generation using a local systems table with fallback to Inventory and persist fetched system details for future use.

New Features:
- Add a query helper to retrieve system details for playbooks from the local systems table.
- Populate ansible_host for a seeded test system to support playbook host resolution.

Bug Fixes:
- Ensure missing systems are filtered from playbook issues and hosts when strict mode is disabled, and UNKNOWN_SYSTEM errors propagate correctly when enabled.

Enhancements:
- Update system resolution in the generator controller to prefer locally stored system details, only calling Inventory for missing systems and backfilling them into the database.
- Refine host mapping logic to base hosts on successfully resolved systems and drop issues with no valid hosts in non-strict mode.

Tests:
- Add unit tests for resolveSystems to cover local DB usage, Inventory fallback, strict vs non-strict handling, and issue filtering behavior.
- Add unit tests for getSystemDetailsForPlaybook to verify DB lookups, ansible_host mapping, and handling of missing or empty inputs.

### DIFF
--- a/src/generator/generator.controller.js
+++ b/src/generator/generator.controller.js
@@ -5,7 +5,9 @@ const P = require('bluebird');
 const etag = require('etag');
 
 const errors = require('../errors');
+const queries = require('../remediations/remediations.queries');
 const inventory = require('../connectors/inventory');
+const { storeSystemDetails } = require('../remediations/controller.write');
 const templates = require('../templates/static');
 const SpecialPlay = require('./plays/SpecialPlay');
 const format = require('./format');
@@ -126,17 +128,17 @@ exports.systemToHost = function (system) {
 };
 
 /**
- * Resolves system IDs to hostnames by fetching system details from Inventory.
+ * Resolves system IDs to hostnames by first checking the local systems table,
+ * then falling back to Inventory for any missing systems.
  * Adds a `hosts` property to each issue containing the resolved hostnames.
  *
  * @param {Array} issues - Array of issue objects, each containing a `systems` array of system IDs
  * @param {boolean} strict - Controls error handling for missing systems:
- *   - true (default): Throws UNKNOWN_SYSTEM error if any system ID is not found in Inventory
- *   - false: Gracefully handles missing systems by filtering them out and removing empty issues
+ *   - true (default): Throws UNKNOWN_SYSTEM error if any system ID is not found
+ *   - false: Filters out missing systems from hosts and removes issues with no valid systems/hosts
  *
- * Returns the issues array with `hosts` added to each issue
- * Or throws UNKNOWN_SYSTEM error if strict=true and some systems weren't found in Inventory.
- * When strict=false, missing systems are filtered out and issues with no remaining systems are removed.
+ * Returns the issues array with `hosts` added. Issues with no valid hosts are removed (strict=false only).
+ * Systems fetched from Inventory are stored locally for future lookups.
  */
 exports.resolveSystems = async function (issues, strict = true) {
     trace.enter('generator.controller.resolveSystems');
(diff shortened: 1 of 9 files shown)


src/remediations/remediations.queries.js
```diff
@@ -955,43 +926,29 @@ exports.getPlanSystemsDetails = async function (inventoryIds, chunkSize = 50) {
...
             result[system.id] = {
+                id: system.id,
                 hostname: system.hostname,
                 ansible_hostname: system.ansible_hostname,
                 display_name: system.display_name
             };
         });
     }
 
-    // For any systems we still couldn't find details for, set default values
```
marleystipich2, line 994: We can remove this block because nothing relied on `null` placeholders for systems that weren't found in the `systems` table or Inventory. I added a strict parameter to `getPlanSystemsDetails` so when strict is false, not found systems are omitted from the result and when strict is true, we  throw an `UNKNOWN_SYSTEM` error. That covers playbook generation (`resolveSystems` passes `strict` through) and FiFi (uses the default `strict=false` and already falls back to the dispatcher host name when a system isn't in the lookup result). The placeholders were redundant before and everything should work the same without this.

src/remediations/remediations.queries.js
```diff
@@ -757,26 +758,13 @@ exports.getIssueSystems = function (id, tenant_org_id, created_by, issueId) {
...
-        ansible_hostname: system.ansible_host || null
-    }));
-
-    if (remediationSystems.length > 0) {
-        await db.systems.bulkCreate(remediationSystems, {
-            updateOnDuplicate: ['hostname', 'display_name', 'ansible_hostname', 'updated_at']
-        });
-    }
+    const inventoryData = await inventory.getSystemDetailsBatch(missingIds, refresh, 2, strict);
+    await storeSystemDetails(inventoryData);
```
marleystipich2, line 767: We already do the work to store system details in the `storeSystemDetails` function so we should just be able to call that function here. So that we only have one place in the code that we call to store system info in the local `systems` table. 

The only difference is that storeSystemDetails does `ignoreDuplicates: true` and the old `fetch_missing_system_data` used to do `updateOnDuplicate`. But this should be fine because `fetch_missing_system_data` only runs for missing system IDs anyway.

src/remediations/remediations.queries.js
```diff
@@ -965,43 +936,29 @@ exports.getPlanSystemsDetails = async function (inventoryIds, chunkSize = 50) {
...
+            where: { id: chunk },
             raw: true
         });
 
         // Add systems to result object
         systems.forEach(system => {
             result[system.id] = {
+                id: system.id,
                 hostname: system.hostname,
                 ansible_hostname: system.ansible_hostname,
```
rexwhite: Make this property `ansible_host` instead, so that it's consistent with the rest of the code (and with the inventory field name).

src/remediations/remediations.queries.js
rexwhite: Since we're doing all this work addressing scalability I feel I should point out that this approach has a hard scale limit.  System identifiers are 36 character UUIDs.  We currently limit plans to 100 *distinct* systems, which is only 3,600 bytes, but there *is* a SQL input buffer size limit and if the list of systems was large enough you would overflow the input buffer.  To get around this you could either:

1. Chunk the lookups and accumulate the list of known systems, then calculate the difference.  

2. A better approach would be to create a temp table, insert the list of inventory ids into the temp table, then do a `SELECT temp.id FROM temp LEFT JOIN systems ON temp.id = systems.id WHERE systems.id IS NULL`.  This way you let the db "do the math", only get back the list of missing systems and don't have to consume memory with a list of known systems that you immediately discard. 

3. To avoid the complexity of temp tables, a hybrid approach would be to just SELECT from a chunked array of values LEFT JOINed on the systems table, and accumulate the unknown systems per chunk.

None of these are *strictly* necessary, but you *have* introduced a hard scale limit here.

src/remediations/remediations.queries.js
```diff
@@ -965,43 +936,29 @@ exports.getPlanSystemsDetails = async function (inventoryIds, chunkSize = 50) {
     const missingIds = _.difference(inventoryIds, existingIds);
 
     // Fetch missing system details from inventory service and store them
-    await fetch_missing_system_data(missingIds);
+    await fetch_missing_system_data(missingIds, { refresh, strict });
 
     const result = {};
 
     // Process in configurable chunks
```
rexwhite: Interesting that you chunk the retrieval of system details here, but not the discovery of missing systems above...

src/remediations/remediations.queries.js
```diff
@@ -930,27 +920,8 @@ exports.getPlanSystems = async function (
...
- * //   },
- * //   '00000000-0000-0000-0000-000000000000': { 
- * //     hostname: '00000000-0000-0000-0000-000000000000', 
- * //     ansible_hostname: null, 
- * //     display_name: null
- * //   }
- * // }
  */
-exports.getPlanSystemsDetails = async function (inventoryIds, chunkSize = 50) {
+exports.getPlanSystemsDetails = async function (inventoryIds, chunkSize = 50, refresh = false, strict = false) {
```
rexwhite: `chunkSize=50` should be the *last* parameter, otherwise you'll have to explicitly specify it if you want to use `refresh` or `strict`.  In fact, the order should probably be: `(inventoryIds, strict = false, refresh = false, chunkSize = 50)`

src/generator/generator.controller.js
```diff
@@ -147,39 +147,27 @@ exports.resolveSystems = async function (issues, strict = true) {
...
-    if (!strict) {
-        trace.event('Remove systems for which we have no inventory entry...');
-        _.forEach(issues, issue => issue.systems = issue.systems.filter((id) => {
-            // eslint-disable-next-line security/detect-object-injection
-            return (systems.hasOwnProperty(id));
-        }));
+    let systems;
+    try {
+        systems = await queries.getPlanSystemsDetails(systemIds, 50, true, strict);
+    } catch (e) {
```
rexwhite, line 149: This is overly broad - if *any* throws because of any sort of error (e.g. the database is down) this will catch it and wrap it in an "unknown system(s)" error.

marleystipich2: This re-throws the original error `throw e;` but the `probes.failedGeneration('unknown system(s)')` line will log it as `unknown system(s)` which still would be confusing.. I can update it to only log `unknown system(s)` if it's an `UNKNOWN_SYSTEM` error

src/remediations/remediations.queries.js
marleystipich2: I checked which endpoints use the `getPlanSystemsDetails` function and actually all of the endpoints that use this function are read endpoints so none of them will hit the 100 distinct systems limit that we put on a couple of the write endpoints.. And the limit is 100 distinct systems **per request** so the plan can have a lot more systems than that so `inventoryIds` can definitely be pretty big 😬 I updated this to process the `inventoryIds` in chunks.

src/remediations/remediations.queries.js
marleystipich2: I updated the other one to process it in chunks too. This was how the function worked before and I'm not sure why it was set up like this 🤔

rexwhite (approved):
Looks good 👍🏻

PR #1182, opened 2026-06-22
RHINENG-25798 RHINENG-25799: Adjust formatting for expires_at in list and get

marleystipich2:
The `null` fallback on `list` was removed so list and get now use the same `expires_at` logic in the format functions. The `_.isDate()` pass-through is intentional because the query functions **should** return Date objects (`get` responses might return ISO strings after `JSON.parse` but that should still work fine too). `expires_at` is always computed in SQL by using the plan `updated_at` time and `last_run_at` time and is required/non-nullable in the OpenAPI spec. Explicit handling for null/undefined isn't needed here because that would basically be an internal bug that we need to fix on our end. 

## Summary by Sourcery

Bug Fixes:
- Prevent improper conversion of non-Date expires_at values when formatting remediation list and detail responses.


rexwhite (approved):
Meh... weird, but I see your point. 😜

PR #1186, opened 2026-06-23
RHINENG-25798: Add expires_at to example in api spec

marleystipich2:
## Summary by Sourcery

Documentation:
- Document the expires_at field in the sample payload for this API endpoint in the OpenAPI spec.


rexwhite (approved):
👍🏻

PR #1189, opened 2026-06-23
RHINENG-26252: Improve performance for GET /remediations/:id/connection_status

marleystipich2:
## Summary by Sourcery

Optimize remediation connection status retrieval by reusing a shared query for distinct system IDs and lightweight plan existence checks.

Enhancements:
- Introduce a dedicated query to fetch distinct system IDs for a remediation plan and reuse it in both query and controller layers.
- Rename and repurpose the remediation existence check to a lighter-weight plan existence query used by multiple endpoints.

### DIFF
--- a/src/remediations/controller.fifi_2.js
+++ b/src/remediations/controller.fifi_2.js
@@ -21,29 +21,21 @@ exports.connection_status = errors.async(async function (req, res) {
     const remediationId = req.params.id;
     const tenantOrgId = req.user.tenant_org_id;
     const username = req.user.username;
-    const exclude = req.body.exclude || [];
 
     //----------------------------------------------------------------
-    // fetch remediation
+    // verify remediation exists
     //----------------------------------------------------------------
-    const remediation = await queries.get(remediationId, tenantOrgId, username);
+    const remediation = await queries.checkPlanExistence(remediationId, tenantOrgId, username);
 
     if (!remediation) {
         // 404 if remediation not found
         return notFound(res);
     }
 
-    //--------------------------------------------------------------
-    // Extract unique, sorted list of system_ids from remediation
-    //--------------------------------------------------------------
-    const systemIds = [
-        ... new Set(
-            _(remediation.issues)
-                .flatMap('systems')
-                .map('system_id')
-                .value()
-        )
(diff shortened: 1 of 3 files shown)


src/remediations/remediations.queries.js
```diff
@@ -675,7 +675,7 @@ exports.getSummary = async function (id, tenant_org_id, created_by = null) {
 
 // Return `{ id }` when id, tenant_org_id, and created_by match (remediation exists).
 // Return null if no remediation matches (does not exist or wrong scope).
-exports.checkExecutable = async function (id, tenant_org_id, created_by) {
+exports.checkPlanExistence = async function (id, tenant_org_id, created_by) {
```
marleystipich2, line 678: I decided to rename this because this query really is just checking that the plan exists and it'll make more sense when we use this for other endpoints besides the check executable endpoint

src/remediations/controller.fifi_2.js
```diff
@@ -21,29 +21,21 @@ exports.connection_status = errors.async(async function (req, res) {
     const remediationId = req.params.id;
     const tenantOrgId = req.user.tenant_org_id;
     const username = req.user.username;
-    const exclude = req.body.exclude || [];
```
rexwhite, line 24: What moron added this in the first place?!...

Oh... wait... it was *me* - *I'm* the moron that copy pasted this useless line of code... 🙄

src/remediations/controller.fifi_2.js
```diff
@@ -21,29 +21,21 @@ exports.connection_status = errors.async(async function (req, res) {
...
-            _(remediation.issues)
-                .flatMap('systems')
-                .map('system_id')
-                .value()
-        )
-    ].sort();
+    //----------------------------------------------------------------
+    // fetch distinct system ids for the plan
+    //----------------------------------------------------------------
+    const systemIds = await queries.getPlanSystemIds(remediationId, tenantOrgId, username);
```
rexwhite, line 38: Ahh... I see what you're doing here!  Good job - this is much cleaner!

rexwhite (approved):
🤠👍🏻

PR #1190, opened 2026-06-23
RHINENG-26265: Improve performance for GET /remediation/:id/systems

marleystipich2:
## Summary by Sourcery

Rename and reuse a lightweight remediation existence check to improve performance and clarify intent for remediation-related endpoints.

Bug Fixes:
- Ensure remediation system lookups return 404 when the remediation plan does not exist.

Enhancements:
- Introduce a shared remediation plan existence query and use it across controllers to avoid unnecessary data loading.

### DIFF
--- a/src/remediations/remediations.queries.js
+++ b/src/remediations/remediations.queries.js
@@ -673,16 +673,17 @@ exports.getSummary = async function (id, tenant_org_id, created_by = null) {
     };
 };
 
-// Return `{ id }` when id, tenant_org_id, and created_by match (remediation exists).
+// Return `{ id }` when a remediation matches id and tenant_org_id.
+// created_by is null for service accounts (org-wide access); otherwise filter to that user.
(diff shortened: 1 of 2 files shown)


src/remediations/remediations.queries.js
```diff
@@ -675,7 +675,7 @@ exports.getSummary = async function (id, tenant_org_id, created_by = null) {
 
 // Return `{ id }` when id, tenant_org_id, and created_by match (remediation exists).
 // Return null if no remediation matches (does not exist or wrong scope).
-exports.checkExecutable = async function (id, tenant_org_id, created_by) {
+exports.checkPlanExistence = async function (id, tenant_org_id, created_by) {
```
marleystipich2, line 679: I decided to rename this because this query really is just checking that the plan exists and it'll make more sense when we use this for other endpoints besides the check executable endpoint

rexwhite (approved):
👍🏻

PR #1205, opened 2026-07-06
RHINENG-22737: Block execution of overly large plans

marleystipich2:
## Summary by Sourcery

Enforce configurable size limits on remediation plan execution and update related tests and configuration.

New Features:
- Introduce configurable maximum systems and issues limits for remediation plan playbook execution, with a feature-flag-based bypass option.

Enhancements:
- Apply plan size validation before executing FiFi playbook runs to prevent overly large remediation plans from running.

Build:
- Add plan size limit configuration (systems, issues, bypass feature flag) to the application config.

Tests:
- Add integration tests covering plan size limit enforcement, bypass behavior via feature flag, and adjust large-remediation creation test to use 19k systems with an extended timeout.

### DIFF
--- a/src/config/__snapshots__/config.unit.js.snap
+++ b/src/config/__snapshots__/config.unit.js.snap
@@ -130,6 +130,11 @@ exports[`Configuration Verify app-common-js function 1`] = `
     "base": "/api/remediations",
     "prefix": "/api",
   },
+  "planLimits": {
+    "bypassFeatureFlag": "remediations.plan-limits-bypass",
+    "maxIssues": 20000,
+    "maxSystems": 20000,
+  },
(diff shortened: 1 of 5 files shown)


src/config/__snapshots__/config.unit.js.snap
```diff
@@ -130,6 +130,11 @@ exports[`Configuration Verify app-common-js function 1`] = `
     "base": "/api/remediations",
     "prefix": "/api",
   },
+  "planLimits": {
```
marleystipich2, line 133: I used `20000` as a generic really high limit lol but we can use a higher or lower number. We can also adjust this number overtime if we realize this doesn't work for us.

PR #1210, opened 2026-07-08
RHINENG-28559: Replace uuid dependency with crypto.randomUUID (foreman-3.18)

marleystipich2:
## Summary by Sourcery

Replace usage of the external uuid library with Node.js crypto.randomUUID across remediations, middleware, connectors, and tests, while keeping playbook run ID generation testable.

Enhancements:
- Introduce a generatePlaybookRunId helper in fifi_2 to encapsulate playbook run ID creation and make it easier to stub in tests.
- Simplify request ID and connector mocks to rely on crypto.randomUUID instead of a third-party UUID generator.

Build:
- Remove the uuid dependency from package.json now that UUID generation uses the built-in crypto module.

Tests:
- Update unit and integration tests to stub or generate IDs using crypto.randomUUID and the new generatePlaybookRunId helper instead of uuid.v4.

Chores:
- Clean up unused uuid imports from seeders and inventory utilities that no longer rely on the library.

### DIFF
--- a/db/seeders/20240326162729-scalability.js
+++ b/db/seeders/20240326162729-scalability.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const _ = require('lodash');
-const UUID = require('uuid');
 const { account_number, tenant_org_id, username: created_by } = require('../../src/connectors/users/mock').MOCK_USERS.fifi;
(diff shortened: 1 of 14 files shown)


vkrizan (approved):
Thank you!

PR #1211, opened 2026-07-08
RHINENG-28559: Replace uuid dependency with crypto.randomUUID (foreman-3.16)

marleystipich2:
## Summary by Sourcery

Replace usage of the external uuid library with Node.js built-in crypto.randomUUID across remediations, middleware, connectors, and tests, and expose a dedicated playbook run ID generator for easier stubbing in tests.

Enhancements:
- Introduce a generatePlaybookRunId helper in fifi_2 to encapsulate playbook run ID creation and make it easier to stub in tests.
- Refactor receptor connector mock and request ID middleware to rely on crypto.randomUUID instead of the uuid package.

Build:
- Remove the uuid dependency from package.json now that ID generation uses the built-in crypto module.

Tests:
- Update integration and unit tests to stub generatePlaybookRunId and to use crypto.randomUUID-generated IDs instead of the uuid library.

### DIFF
--- a/db/seeders/20240326162729-scalability.js
+++ b/db/seeders/20240326162729-scalability.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const _ = require('lodash');
-const UUID = require('uuid');
 const { account_number, tenant_org_id, username: created_by } = require('../../src/connectors/users/mock').MOCK_USERS.fifi;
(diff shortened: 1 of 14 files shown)


vkrizan (approved):
Thank you!

PR #1236, opened 2026-07-21
RHINENG-28446: Clear cache when updating remediation plan

marleystipich2:
## Summary by Sourcery

Invalidate remediation cache in Redis when remediation plans or issues are updated or deleted.

Enhancements:
- Introduce a reusable helper to clear remediation cache entries when a remediation or its issues are patched or removed.

Tests:
- Add integration test verifying remediation cache invalidation on remediation patch operations.

### DIFF
--- a/src/remediations/controller.write.js
+++ b/src/remediations/controller.write.js
@@ -8,12 +8,24 @@ const config = require('../config');
 const errors = require('../errors');
 const issues = require('../issues');
 const db = require('../db');
+const cache = require('../cache');
 const format = require('./remediations.format');
 const inventory = require('../connectors/inventory');
 const identifiers = require('../util/identifiers');
 const disambiguator = require('../resolutions/disambiguator');
+const log = require('../util/log');
 
 const notFound = res => res.status(404).json();
+
+async function clearRemediationsCache (id) {
+    if (config.redis.enabled) {
+        try {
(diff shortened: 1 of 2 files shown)


src/remediations/controller.write.js
```diff
@@ -8,13 +8,25 @@ const config = require('../config');
...
+const cache = require('../cache');
 const format = require('./remediations.format');
 const inventory = require('../connectors/inventory');
 const identifiers = require('../util/identifiers');
 const disambiguator = require('../resolutions/disambiguator');
+const log = require('../util/log');
 
 const notFound = res => res.status(404).json();
 
+async function clearRemediationsCache (id) {
```
marleystipich2, line 20: We have a similar function in `remediations.queries.js` for clearing the org config cache:
```
exports.clearOrgConfigCache = async function (tenant_org_id) {
    if (config.redis.enabled && cache.get().status === 'ready') {
        try {
            await cache.get().del(ORG_CONFIG_CACHE_KEY(tenant_org_id));
        } catch (err) {
            log.warn(err, 'failed to clear org config retention cache');
        }
    }
};
```
But `remediations.queries.js` seems like a weird place to put a cache clearing function because that file is supposed to be for DB queries.. So I put this new `clearRemediationsCache` helper function in `src/remediations/controller.write.js` since that's where we're actually using it 🤷‍♀️

PR #1243, opened 2026-07-27
chore(deps): refresh rpm lockfiles [SECURITY]

marleystipich2:
I noticed the Konflux lockfile refresh PRs haven't been updating rpms.lock.yaml so I looked into it and noticed that things are pretty out of sync:
- rpms.lock.yaml was last updated on May 20
- Base image in the Dockerfile has been updated several times since then
- `BASE_IMAGE` in the Makefile hasn't been updated since Oct 2025
- `ubi.repo` was added in November 2025 but has never been updated

This updates the `BASE_IMAGE` in the Makefile so that it matches the version that we're using in the Dockerfile. Also refreshed the ubi.repo file to keep it in sync with the base image in the Dockerfile as well. The most important thing this PR does is correctly refresh `rpms.lock.yaml` since it hasn't been updated in ~10 weeks. 

I think we should look into how MintMaker/Renovate are configured to make sure the lockfile refresh stuff is updating the correct files so things don't get out of sync without us realizing it. 

## Summary by Sourcery

Build:
- Update default UBI9 minimal base image tag used for container builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # Default value for BASE_IMAGE
-BASE_IMAGE ?= registry.access.redhat.com/ubi9/ubi-minimal:1784092978
+BASE_IMAGE ?= registry.access.redhat.com/ubi9/ubi-minimal:1785339196
 # Default value for CONTAINERFILE
 CONTAINERFILE ?= Dockerfile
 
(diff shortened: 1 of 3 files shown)


marleystipich2:
@vkrizan You guys might want to make similar changes on the foreman branches.. It seems like Konflux isn't refreshing the `rpm.lock.yaml` (still need to figure out what's going on with that) but ours hasn't been updated in ~10 weeks. There were some other things that were out of sync too but I added more info in the description of this PR.

marleystipich2:
Did all of this in this other PR that was merged already: https://github.com/RedHatInsights/insights-remediations/pull/1278
Closing this out.

PR #1244, opened 2026-07-28
RHINENG-29341: Adjust renovate config for master branch

marleystipich2:
These are the renovate config changes made in this PR:
- Removed the old “disable all minor/patch” rule that blocked every minor or patch update. This rule technically cut down PR noise but it could also skip a security bump that landed as minor or patch.
- Disabled routine/maintenance npm bumps so we only get Konflux PRs for security-related npm updates.
- When the Dockerfile base image is bumped, refresh the RPM lockfile in the same PR.
- Disabled routine RPM lockfile refresh PRs (we were getting lockfile PRs for every RPM update, including non-security ones).
- Still allow standalone RPM lockfile refresh PRs when they are security-related.

## Summary by Sourcery

Build:
- Update Renovate bot configuration for the master branch.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,6 +3,7 @@
   "extends": [
     "github>RedHatInsights/konflux-pipelines//renovate/foreman_satellite/renovate.json"
   ],
+  "baseBranchPatterns": ["master", "main", "/^foreman-.*$/", "/^SATELLITE-.*$/"],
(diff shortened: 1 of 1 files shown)


rexwhite (approved):
Oooh... this looks like it should be much better.  Thanks for taking care of this!

PR #1284, opened 2026-08-06
[RHCLOUD-49129] Upgrade kessel version

astrozzc:
https://redhat.atlassian.net/browse/RHCLOUD-49129
RBAC require explicit parameter to bypass the permission check when fetching default workspace

## Summary by Sourcery

Build:
- Update the @project-kessel/kessel-sdk package dependency from ^3.2.0 to ^3.8.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@godaddy/terminus": "^4.4.1",
-        "@project-kessel/kessel-sdk": "^3.2.0",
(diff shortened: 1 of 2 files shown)


marleystipich2 (approved):
LGTM! 👍 
Note to self so I remember why this was needed: Without this bump, our older Kessel SDK doesn’t send `with_ancestry` at all so RBAC treats the call as `with_ancestry=false`. Bump `@project-kessel/kessel-sdk` to 3.8.0 so `fetchDefaultWorkspace` sends `with_ancestry=true` by default and our default-workspace lookup keeps working.

